PR #8, opened 2022-02-14
Add configurable keybinds

ByMartrixx:
Un-hardcodes every used key, adds menu to configure key binds to any key, located in `File > Configure Keybinds...`

image
image
image


OroArmor (approved):

LambdAurora (approved):

ByMartrixx:
Added support for changing the modifier keys used in each keybind
image
image
image

OroArmor (approved):

enigma-swing/src/main/java/cuchaz/enigma/gui/config/KeyBindsConfig.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private static ConfigSection runningCfg;
+
+    static {
+        KeyBindsConfig.snapshotConfig();
+    }
+
+    // Save the current configuration state so consistent keybinds can be provided
+    // for keybinds that are not configurable during runtime.
+    public static void snapshotConfig() {
```
2xsaiko, line 23: Hmmm, is it not possible to change all these in the UI immediately when set through the options? Need to take a closer look later

ByMartrixx: I managed to make it work without issues

2xsaiko (approved):
LGTM! Very nice.

PR #16, opened 2022-11-05
Update Tests, Remove a lots of guava calls

kadir1243:
This pull contains:
- Remove a lots of guava calls(It Should decrease size of shadowed jars)
- Update Tests to Junit 5
- Change Some Throw Statements (I hope this will not broke anything)
- Update dependencies
- Change Local Variable Name Fixer system to detect compiled java version and not apply fixes if not needed for java version

### DIFF
--- a/enigma/src/main/java/cuchaz/enigma/translation/mapping/serde/srg/SrgMappingsWriter.java
+++ b/enigma/src/main/java/cuchaz/enigma/translation/mapping/serde/srg/SrgMappingsWriter.java
@@ -1,14 +1,13 @@
 package cuchaz.enigma.translation.mapping.serde.srg;
 
-import com.google.common.collect.Lists;
 import cuchaz.enigma.ProgressListener;
 import cuchaz.enigma.translation.MappingTranslator;
 import cuchaz.enigma.translation.Translator;
 import cuchaz.enigma.translation.mapping.EntryMapping;
 import cuchaz.enigma.translation.mapping.MappingDelta;
-import cuchaz.enigma.translation.mapping.serde.MappingSaveParameters;
 import cuchaz.enigma.translation.mapping.VoidEntryResolver;
 import cuchaz.enigma.translation.mapping.serde.LfPrintWriter;
+import cuchaz.enigma.translation.mapping.serde.MappingSaveParameters;
 import cuchaz.enigma.translation.mapping.serde.MappingsWriter;
 import cuchaz.enigma.translation.mapping.tree.EntryTree;
 import cuchaz.enigma.translation.mapping.tree.EntryTreeNode;
@@ -26,6 +25,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
+import java.util.stream.StreamSupport;
 
 public enum SrgMappingsWriter implements MappingsWriter {
 	INSTANCE;
@@ -43,7 +43,7 @@ public enum SrgMappingsWriter implements MappingsWriter {
 		List<String> fieldLines = new ArrayList<>();
 		List<String> methodLines = new ArrayList<>();
 
-		List<? extends Entry<?>> rootEntries = Lists.newArrayList(mappings).stream()
+		List<? extends Entry<?>> rootEntries = StreamSupport.stream(mappings.spliterator(), false)
 				.map(EntryTreeNode::getEntry)
 				.toList();
 		progress.init(rootEntries.size(), I18n.translate("progress.mappings.srg_file.generating"));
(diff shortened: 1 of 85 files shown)


enigma/src/main/java/cuchaz/enigma/translation/mapping/serde/tinyv2/TinyV2Reader.java
```diff
@@ -111,19 +110,17 @@ private EntryTree<EntryMapping> read(Path path, List<String> lines, ProgressList
 
 						if (state.get(IN_CLASS)) {
 							switch (parts[0]) {
-								case "m": // method
+								case "m" -> { // method
 									state.set(IN_METHOD);
 									holds[IN_METHOD] = parseMethod(holds[IN_CLASS], parts, escapeNames);
-									break;
```
NoComment1105, line 117: Why have all the breaks been removed from this switch? Have you verified they're unnecessary?

enigma/src/main/java/cuchaz/enigma/translation/representation/TypeDescriptor.java
```diff
@@ -43,7 +42,7 @@ public TypeDescriptor(String desc) {
 
 	public static String parseFirst(String in) {
 
-		if (in == null || in.length() <= 0) {
+		if (in == null || in.length() == 0) {
```
NoComment1105, line 45: Have you verified this and other locations don't carry the risk of being less than 0?

NoComment1105 (changes requested):
Looking mostly good, there's a lot of inconsistency around removal of some wildcard imports but conversion to wildcard imports that could do with being looked at if possible. I'm not an Enigma dev but I like what I see :P

enigma/src/main/java/cuchaz/enigma/translation/mapping/serde/tinyv2/TinyV2Reader.java
Geolykt: I believe they are not required with the new switch-case syntax

NoComment1105: Interesting, I'm not convinced

enigma/src/main/java/cuchaz/enigma/translation/representation/TypeDescriptor.java
Geolykt: Have you ever seen a string that is -1 codepoints long? It's fair to assume that in the event of there being an integer overflow that we have larger worries (if we were to assume that the JVM supports arrays of that size - which I don't think it does)

enigma/src/main/java/cuchaz/enigma/translation/mapping/serde/srg/SrgMappingsWriter.java
```diff
@@ -110,8 +110,8 @@ private String describeField(FieldEntry entry) {
 		return entry.getParent().getFullName() + "/" + entry.getName();
 	}
 
-	private Collection<Entry<?>> sorted(Iterable<? extends Entry<?>> iterable) {
-		ArrayList<Entry<?>> sorted = Lists.newArrayList(iterable);
+	private Collection<Entry<?>> sorted(Collection<? extends Entry<?>> iterable) {
+		List<Entry<?>> sorted = new ArrayList<>(iterable);
```
ByMartrixx, line 114:
```suggestion
	private Collection<Entry<?>> sorted(Collection<? extends Entry<?>> entries) {
		List<Entry<?>> sorted = new ArrayList<>(entries);
```

enigma/src/main/java/cuchaz/enigma/translation/mapping/serde/tinyv2/TinyV2Reader.java
ByMartrixx: Yeah, switch expressions don't need breaks

enigma/src/main/java/cuchaz/enigma/utils/Utils.java
```diff
@@ -51,6 +43,19 @@ public static void delete(Path path) throws IOException {
...
+        Path name = path.getFileName();
+
+        // null for empty paths and root-only paths
+        if (name == null) {
+            return "";
+        }
+
+        String fileName = name.toString();
+        int dotIndex = fileName.lastIndexOf('.');
+        return dotIndex == -1 ? "" : fileName.substring(dotIndex + 1);
```
ByMartrixx, line 56: I don't think this handles files or directories starting with a dot and without an extension

ByMartrixx (changes requested):
Just a couple things:
- the tests don't run with `gradlew test`.
- A lot of changes were just 5 imports from the same package changed to a `*` import, which is quite unnecessary imo.

Looks good aside from that and the comments on the files.

kadir1243:
I am going to split this pull(because its a little big) to little pulls

PR #23, opened 2022-12-29
add improved side panel selector and clean up gui

ix0rai:
this pr implements a better selector for side panels and improves a lot of gui code
inspired by, and steals code from, https://github.com/FabricMC/Enigma/pull/420
image

changelog (incomplete):
- add side panel selector
- add translation for "send message" button in multiplayer chat

todo:
- [x] bug: structure panel flashes for a second when a new class is opened
- [x] bug: panels often forget their states when hiding and showing
- [x] bug: pressed side button does not always perfectly reflect selected side panel
- [x] feature: add little "enigma is offline" message to multiplayer windows when enigma is offline
- [x] feature: consistent text box size in "users" panel
- [x] feature: make it less ugly
- [x] feature: title bar above each panel to indicate what's open
- [x] change: make the whole registration system more robust
- [x] change: find a better solution than saiko's insane jlayer rotation code


enigma-swing/src/main/java/cuchaz/enigma/gui/config/UiConfig.java
```diff
@@ -15,9 +16,9 @@ private UiConfig() {
 	}
 
 	// General UI configuration such as localization
-	private static final ConfigContainer ui = ConfigContainer.getOrCreate("enigma/enigmaui");
+	private static final ConfigContainer ui = ConfigContainer.getOrCreate("enigma/ui");
 	// Swing specific configuration such as theming
-	private static final ConfigContainer swing = ConfigContainer.getOrCreate("enigma/enigmaswing");
+	private static final ConfigContainer swing = ConfigContainer.getOrCreate("enigma/swing");
```
ByMartrixx, line 21: These shouldn't be changed, they are paths for config files

enigma-swing/src/main/java/cuchaz/enigma/gui/panels/right/AbstractInheritanceTree.java
```diff
@@ -46,12 +45,10 @@ private void onClick(MouseEvent event) {
...
-			if (node instanceof AbstractClassTreeNode classNode) {
-				gui.getController().navigateTo(new ClassEntry(classNode.getClassName()));
-			} else if (node instanceof AbstractMethodTreeNode methodNode) {
-				if (!(methodNode instanceof MethodInheritanceTreeNode inheritanceNode) || inheritanceNode.isImplemented()) {
-					gui.getController().navigateTo(methodNode.getMethodEntry());
-				}
+			if (node instanceof ClassInheritanceTreeNode classNode) {
+				gui.getController().navigateTo(classNode.getClassEntry());
+			} else if (node instanceof MethodInheritanceTreeNode methodNode && methodNode.isImplemented()) {
+				gui.getController().navigateTo(methodNode.getMethodEntry());
```
ByMartrixx, line 51: Looks like there was an improper merge here, the class one should be `instanceof AbstractClassNode` and the method one `instanceof AbstractMethodTreeNode`, which should be always navigable to as opposed to an `InheritanceNode`, which should only navigable if it's implemented

ByMartrixx (commented):
Did a quick review, gonna check everything once the pr is done

OroArmor (approved):

enigma-swing/src/main/java/cuchaz/enigma/gui/panels/right/RightRotatedLayerUI.java
```diff
@@ -0,0 +1,362 @@
...
+import java.awt.Point;
+import java.awt.event.MouseEvent;
+import java.awt.event.MouseListener;
+import java.awt.event.MouseMotionListener;
+import java.awt.event.MouseWheelEvent;
+import java.awt.event.MouseWheelListener;
+
+/**
+ * Originally by 2xsaiko.
+ * todo look into replacing this with something simpler
```
2xsaiko, line 22: good luck lmao
(I will actually be impressed if you can find something simpler that isn't extremely janky)

ix0rai: glad I have your blessing!
if all else fails I'm just going to hardcore it to 90° rotations, counterclockwise and clockwise.

ix0rai: I've been defeated, stackoverflow has nothing for me
your code stays!

enigma-swing/src/main/java/cuchaz/enigma/gui/config/UiConfig.java
```diff
@@ -6,10 +6,66 @@
 
 import cuchaz.enigma.config.ConfigContainer;
 import cuchaz.enigma.config.ConfigSection;
+import cuchaz.enigma.gui.panels.right.RightPanel;
 import cuchaz.enigma.gui.util.ScaleUtil;
 import cuchaz.enigma.utils.I18n;
 
 public final class UiConfig {
+	// sections
+	public static final String MAIN_WINDOW = "Main Window";
```
ByMartrixx, line 15: Is there any reason for these fields to be public?

enigma-swing/src/main/java/cuchaz/enigma/gui/panels/right/RightPanel.java
```diff
@@ -0,0 +1,75 @@
...
+		String translatedTitle = this.titleProvider.get();
+        this.button.setText(translatedTitle);
+		this.title.setText(translatedTitle);
+    }
+
+    public JToggleButton getButton() {
+        return this.button;
+    }
+
+    public static void registerPanel(RightPanel panel) {
```
ByMartrixx, line 50: `addPanel` would be better imo. Or rather, seeing that using each instance of a right panel needs a cast, using a `Class<?>` as key would be an alternative, so you would have this method to associate each panel class to an instance, and a generic method to retrieve a panel instance based on the given `Class<?>`, which would probably end up in a single unchecked cast warning

enigma-swing/src/main/java/cuchaz/enigma/gui/panels/right/RightPanel.java
```diff
@@ -0,0 +1,75 @@
...
+
+import javax.swing.JLabel;
+import javax.swing.JPanel;
+import javax.swing.JToggleButton;
+import java.awt.BorderLayout;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Supplier;
+
+public abstract class RightPanel extends JPanel {
```
ByMartrixx, line 14: Found a bug. Seems like the width for the panels reset to 0 when changing between panels under some conditions. Reproduction steps:
1. Launch enigma
2. Open a right panel if one isn't there already
3. Click on the button for the selected panel to close it
4. Open a different panel
5. Click on the button for the previous panel (same on as step 4)
6. A panel won't appear, until you try using the 3 dots to resize it

enigma-swing/src/main/java/cuchaz/enigma/gui/panels/right/UsersPanel.java
```diff
@@ -0,0 +1,38 @@
...
+
+import javax.swing.JPanel;
+import javax.swing.JScrollPane;
+import java.awt.BorderLayout;
+
+public class UsersPanel extends MultiplayerOnlyRightPanel {
+	private final JPanel topPanel;
+
+    public UsersPanel(Gui gui) {
+        super(gui);
```
ByMartrixx, line 13: This panel doesn't have a title

enigma-swing/src/main/java/cuchaz/enigma/gui/elements/EditorTabbedPane.java
```diff
@@ -80,7 +81,9 @@ public void onTitleChanged(EditorPanel editor, String title) {
 
 		if (editorPanel != null && activeEditor != editorPanel) {
 			this.openFiles.setSelectedComponent(this.editors.get(entry).getUi());
-			this.gui.showStructure(editorPanel);
+			if (gui.getRightPanel().getId().equals(RightPanel.Type.STRUCTURE)) {
+				this.gui.showStructure(editorPanel);
```
ByMartrixx, line 85: This should be invoked on any conditions to update the panel, even if it's not the current one. Otherwise, changing the selected class when it isn't open keeps the structure of the previous class.
Following that, the method should be renamed or moved to `updateStructure`

enigma-swing/src/main/java/cuchaz/enigma/gui/Gui.java
```diff
@@ -104,18 +111,34 @@ public class Gui {
 	public SearchDialog searchDialog;
 
 	public Gui(EnigmaProfile profile, Set<EditableType> editableTypes) {
+		// right panels
+		// top panels
+		RightPanel.registerPanel(new CallsTree(this));
+		RightPanel.registerPanel(new ImplementationsTree(this));
+		RightPanel.registerPanel(new InheritanceTree(this));
+		RightPanel.registerPanel(new StructurePanel(this));
```
ByMartrixx, line 119: Order of the panels is inverse here, structure should be the first one

enigma-swing/src/main/java/cuchaz/enigma/gui/Gui.java
```diff
@@ -104,18 +111,34 @@ public class Gui {
...
+
+		// bottom panels
+		RightPanel.registerPanel(new MessagesPanel(this));
+		RightPanel.registerPanel(new UsersPanel(this));
```
ByMartrixx, line 123: Might be worth merging these two into a "Online" tab or similar, just one of the two isn't going to use the whole side on its own

enigma-swing/src/main/java/cuchaz/enigma/gui/Gui.java
```diff
@@ -209,12 +214,56 @@ public void actionPerformed(ActionEvent e) {
...
+	public RightPanel getRightPanel() {
+		return this.rightPanel;
+	}
+
+	/**
+	 * Sets the right panel to the given panel.
+	 * @param id the id of the new panel
+	 * @param updateStateIfCurrent if the provided id is equal to the current id, this parameter determines whether to update the visibility of the panel
+	 */
+	public void setRightPanel(String id, boolean updateStateIfCurrent) {
```
ByMartrixx, line 226: This method is probably to blame for the bug I mentioned in `RightPanel.java`

enigma-swing/src/main/java/cuchaz/enigma/gui/Gui.java
```diff
@@ -209,12 +214,56 @@ public void actionPerformed(ActionEvent e) {
...
+		if (id.equals(this.rightPanel.getId()) && updateStateIfCurrent) {
+			// only save divider location if hiding panel
+			if (this.rightPanel.isVisible()) {
+				UiConfig.setRightPanelDividerLocation(id, this.splitRight.getDividerLocation());
+			}
+
+			// swap visibility
+			this.rightPanel.setVisible(!this.rightPanel.isVisible());
+		} else {
```
ByMartrixx, line 235: This else changes the right panel even if it's the current one when `updateStateIfCurrent` is false

enigma-swing/src/main/java/cuchaz/enigma/gui/Gui.java
```diff
@@ -344,31 +394,44 @@ public void startRename(EditorPanel editor) {
 	}
 
 	public void showStructure(EditorPanel editor) {
```
ByMartrixx, line 396: As I mentioned in `EditorTabbedPane.java`, either this method should be renamed to `updateStructure` or a new method `updateStructure` should be added (which shouldn't change the selected right panel)

enigma-swing/src/main/java/cuchaz/enigma/gui/Gui.java
```diff
@@ -344,31 +394,44 @@ public void startRename(EditorPanel editor) {
...
+	 * @param editor the editor to extract the reference from
+	 */
 	public void showInheritance(EditorPanel editor) {
 		EntryReference<Entry<?>, Entry<?>> cursorReference = editor.getCursorReference();
 		if (cursorReference == null) return;
 
-		this.inheritanceTree.display(cursorReference.entry);
-		tabs.setSelectedIndex(1);
+		this.setRightPanel(RightPanel.Type.INHERITANCE, false);
+		((InheritanceTree) this.getRightPanel()).display(cursorReference.entry);
```
ByMartrixx, line 410: `cursorReference.entry` is Nullable iirc

enigma-swing/src/main/java/cuchaz/enigma/gui/panels/right/MessagesPanel.java
```diff
@@ -0,0 +1,85 @@
...
+public class MessagesPanel extends MultiplayerOnlyRightPanel {
+    private final Gui gui;
+    private final JScrollPane messageScrollPane;
+    private final JTextField pendingMessageBox;
+    private final JButton sendPendingMessageButton;
+    private final JPanel chatPanel;
+
+    public MessagesPanel(Gui gui) {
+        super(gui);
+        this.gui = gui;
```
ByMartrixx, line 24: `gui` should be a protected field of `RightPanel`, instead of a private one for `CallsTree`, `StructurePanel` and `MessagesPanel`, which also shadows a private one on `MultiplayerOnlyRightPanel`

ByMartrixx (changes requested):
Requesting changes since it has a bug that's pretty easy to stumble upon, and gets annoying after a couple times (see my comment `RightPanel.java` for details). Otherwise most stuff looks great

enigma-swing/src/main/java/cuchaz/enigma/gui/elements/JMultiLineToolTip.java
```diff
@@ -13,18 +14,20 @@
 
 /**
  * Implements a multi line tooltip for GUI components
- * Copied from http://www.codeguru.com/java/articles/122.shtml
+ * Copied from <a href="http://www.codeguru.com/java/articles/122.shtml">this CodeGuru article</a>
```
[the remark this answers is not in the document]
ix0rai: will fix!

enigma-swing/src/main/java/cuchaz/enigma/gui/config/UiConfig.java
ix0rai: the entire reason to have them public is the main window constant being used in `Gui`
you know me and my consistency

enigma-swing/src/main/java/cuchaz/enigma/gui/panels/right/RightPanel.java
ix0rai: >  and a generic method to retrieve a panel instance based on the given Class<?>
how would I write this?

enigma-swing/src/main/java/cuchaz/enigma/gui/Gui.java
ix0rai: what do you mean by this?
the previous `if` block runs if the panel is equal to the current one, there's no way for the second block to be triggered unless the panel is being swapped

enigma-swing/src/main/java/cuchaz/enigma/gui/panels/right/RightPanel.java
ByMartrixx: If you have a field `Map<Class<? extends RightPanel>, ? extends RightPanel> panels`, your method would be like this
```java
public static <T extends RightPanel> T getPanel(Class<T> clazz) {
    return (T) panels.get(clazz);
}
```

enigma-swing/src/main/java/cuchaz/enigma/gui/Gui.java
ix0rai: I don't think so
this is the same way the previous code did it, and it never seemed to cause any issues

enigma-swing/src/main/java/cuchaz/enigma/gui/Gui.java
ByMartrixx: There is an and condition, so when `updateStateIfCurrent` is false it doesn't run

enigma-swing/src/main/java/cuchaz/enigma/gui/Gui.java
ByMartrixx: Oh, I didn't see it. It's alright then

enigma-swing/src/main/java/cuchaz/enigma/gui/panels/right/RightPanel.java
ByMartrixx: You might want to throw an exception if the panel doesn't exist, instead of returning null

enigma-swing/src/main/java/cuchaz/enigma/gui/Gui.java
ix0rai: you're right; my mistake

enigma/src/main/resources/lang/en_us.json
```diff
@@ -268,5 +261,14 @@
...
+	"keybind.menu.save": "Save mappings",
+
+	"right_panel.messages.send": "Send",
+    "right_panel.messages.title": "Messages",
+    "right_panel.multiplayer.offline_text": "Enigma is currently running offline.",
+	"right_panel.calls.title": "Calls",
+    "right_panel.structure.title": "Structure",
+    "right_panel.inheritance.title": "Inheritance",
+    "right_panel.implementations.title": "Implementations",
+    "right_panel.users.title": "Users"
```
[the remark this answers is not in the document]
ix0rai: done; I also fixed the indentation according to the new editorconfig

enigma-swing/src/main/java/cuchaz/enigma/gui/panels/right/MultiplayerPanel.java
```diff
@@ -0,0 +1,158 @@
...
+
+	private final Supplier<String> offlineTextProvider = () -> I18n.translate("right_panel.multiplayer.offline_text");
+	private final Supplier<String> usersTitleProvider = () -> I18n.translate("right_panel.multiplayer.users_title");
+	private final Supplier<String> messagesTitleProvider = () -> I18n.translate("right_panel.multiplayer.messages_title");
+	private final Supplier<String> sendButtonTextProvider = () -> I18n.translate("right_panel.multiplayer.send");
+
+	private JPanel panel;
+	private boolean offline;
+
+    public MultiplayerPanel(Gui gui) {
```
ByMartrixx, line 36: The existing menu to start or connect to a server is called "Collab" instead of "Multiplayer", so this should be renamed for consistency. "Multiplayer" doesn't really apply either, since Enigma doesn't have players in first place :>

ByMartrixx (commented):
Just nitpicking

ByMartrixx (approved):
Great job! I'll do some testing and it should be good to go

ix0rai:
*definitely not kicking my feet excitedly right now*

PR #24, opened 2022-12-30
delete legacy ui config code

ix0rai:
exactly what it says on the box!
I highly doubt there's anyone who still has an old enigma ui config who will ever import it to quilt's enigma.


OroArmor (approved):

ix0rai:
extra context: this migration was added mid-2020, so it's pretty much impossible that anyone has one, since they would have needed to import it from fabric's enigma.

ByMartrixx (approved):

Issue #13, opened 2022-08-20
Engima Support (rewriting 90% of the enigma backend)


OroArmor:

ix0rai:
OTHER ENIGMA 2.0 PLANS:

- migrate all packages from `cuchaz` to `org.quiltmc`
- remove access modifier system
- remove most dead code
- evaluate and improve enigma's gui/backend split

PR #26, opened 2022-12-30
Change publication group to org.quiltmc

### LINKED ISSUE #22: Move to org.quiltmc group
Right now, Quilt's fork of enigma doesn't have distinct maven coordinates so it could overlap with other versions. Switching to the `org.quiltmc` group makes it clear that other projects are using this fork (as opposed to one of the many others)

ByMartrixx:
Closes #22

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,7 +30,7 @@ subprojects {
         testImplementation libs.hamcrest
     }
 
-    group = 'cuchaz'
+    group = 'org.quiltmc'
     version = '1.3.2'
 
     var ENV = System.getenv()


ix0rai (approved):
should we consider changing the code package too?

ByMartrixx:
We'd break a lot of stuff if we moved the existing packages now. We might want to wait for #13 too, which would justify moving most if not everything

ix0rai:
good point!
I feel like the QMT migration is a great opportunity to migrate things
I'll leave a comment on that issue

OroArmor:
Id even say wait until enigma is merged into qmt

PR #63, opened 2023-02-19
add support for renaming local classes

SpaceWalkerRS:


tas-pro (approved):

ix0rai (approved):
brilliant!
now to make this visible in the gui :p

SpaceWalkerRS:
should be already, CFR has some trouble decompiling local classes but the other decompilers handle it well

ix0rai:
ah, isn't working for this specific case
I assume it's because this is a java.lang class?

image

SpaceWalkerRS:
That is an anonymous class, not a local class. Local classes are full class definitions inside method bodies, like this case in `net/minecraft/client/Minecraft#main` in 1.3:

Schermafbeelding 2023-02-20 om 00 52 09

ix0rai:
ok, sorry for misinterpreting that

SpaceWalkerRS:
no worries, perhaps I should add some more clarification in the javadocs

ix0rai:
I may be doing something horribly wrong, but it still doesn't appear to be working..?
image

tas-pro:
> I may be doing something horribly wrong, but it still doesn't appear to be working..?

It worked for me using this example: <https://docs.oracle.com/javase/tutorial/java/javaOO/localclasses.html#declaring-local-classes>
image

ix0rai:
still unable to test it, odd
I'll try to figure out why myself

tas-pro:
> still unable to test it, odd I'll try to figure out why myself

Are you using the source from the PR?

ix0rai:
yes, and I've made sure of that
I even tried cloning the source repository
I'm very confused

ByMartrixx:
Is the local prefix needed by the JVM?

ix0rai:
there's nothing there specifically to make the JVM happy, I just followed the convention from the other tests in the package

ByMartrixx (changes requested):
First, I apologize if the comment is a bit too long, but I figured it would be better to explain everything rather than leave details in the air. On to the review, IMHO, this is the wrong way to approach local class detection/mapping.

As far as I know, the JVM doesn't impose any restriction on the binary name for local classes, and obfuscators are free to mangle their name as they like, so a local class without a local prefix would be valid and wouldn't be mappable with this.

That aside, this approach also breaks API on `ClassEntry`s:
- Altough there isn't a clear API-impl separation in Enigma, `ClassEntry` is used pretty much everywhere, and by everything that interacts with Enigma on a code level
- Any application instantiating a `ClassEntry` would crash when updating Enigma, since `new ClassEntry(ClassEntry, String)` no longer exists
- Enigma clients with older versions would expect a different packet format, and would cause errors when receiving any packet containing information about a class.

I'll suggest a different approach instead. In theory, just returning `false` on `EnigmaProject#isAnonymous` (`#isAnonymousOrLocal`) for local classes would allow mapping them.
As for their indentification, half of it is already done: the `EnclosingMethod` attribute is required for local and anonymous classes ([JVMS§4.7.7](https://docs.oracle.com/javase/specs/jvms/se17/html/jvms-4.html#jvms-4.7.7)). To identify whether a class is local or anonymous, you can refer to the `InnerClasses` attribute, which, as specified by [JVMS§4.7.6](https://docs.oracle.com/javase/specs/jvms/se17/html/jvms-4.html#jvms-4.7.6), `inner_name_index` must be 0 for anonymous classes, otherwise a valid index into the `constant_pool` - that means, in ASM `ClassVisitor#visitInnerClass`, anonymous classes will have the `innerName` as null, while local ones won't. To implement this, you can reference `EnclosingMethodIndex` to create a new `JarIndexer` (see 22db343838db9181916045e1530acecc3d4a0389), or implement it in that class. There, you'd store which `ClassEntry`s have a nonnull `innerName` (the value of it is irrelevant), and then you'd check that value from `#isAnonymous`

SpaceWalkerRS:
You bring up some good points. I agree that detecting them based on the name is not the way to go, as the JVM spec does not require names of inner classes to be in any particular form. But that exposes a fundamental flaw in Enigma's current implementation. Enigma assumes internal class names to be in the form `com/example/OuterClassName$InnerName` and uses that to get both the outer class and the inner name (the thing you want to map), but neither piece of information is required by the JVM spec to be encoded in the internal name of an inner class.

A proper solution would make the `ClassEntry`s use the internal names for `getName` (given that the actual inner name does not need to be unique within an outer class) and the inner names could be accessed through `getSimpleName`. This would be a breaking change in the API, but that is unavoidable if you want to fix the problem at its root.

SpaceWalkerRS:
superseded by #66

PR #66, opened 2023-02-28
rework Enigma's handling of inner classes

SpaceWalkerRS:
This PR serves as a replacement for #63 by reworking how Enigma handles inner classes. I made some big changes and have not fully tested it, so it's still very much WIP.

The core issue is how Enigma deduces two pieces of information about an inner class: its outer class and its inner name. Enigma assumes this information is encoded in the class' internal name. By convention this is the case, as inner class names tend to have the format `path/to/OuterClass$InnerName`, but this is not required by the JVM spec, and already breaks down with anonymous and local classes.

This is fixed by taking this information from the inner class attributes in the class file. This allows for proper remapping of all inner and local classes, but also comes with a few drawbacks.
- The `ClassEntry(String)` constructor now becomes risky to use, as it assumes the given class name is from a top level class. Class entries for inner classes should be made with the `ClassEntry(ClassEntry, String, String)` constructor instead.
- None of the mapping formats support inner classes properly. They all assume class names follow convention in one way or another. For example, the tiny formats only deal with a class' internal name, and an inner class' simple name will have to be inferred from that. Meanwhile, the Enigma format only identifies inner classes by their inner name (or rather, the inner name as inferred from the class' internal name) and the class' internal name will have to be inferred from that.

### DIFF
--- a/enigma-server/src/main/java/cuchaz/enigma/network/packet/PacketHelper.java
+++ b/enigma-server/src/main/java/cuchaz/enigma/network/packet/PacketHelper.java
@@ -40,7 +40,13 @@ public class PacketHelper {
 					throw new IOException("Class requires class parent");
 				}
 
-				return new ClassEntry((ClassEntry) parent, name, javadocs);
+				String simpleName;
+				if (input.readBoolean()) {
+					simpleName = input.readBoolean() ? readString(input) : null;
+				} else {
+					simpleName = ClassEntry.getSimpleName(name);
+				}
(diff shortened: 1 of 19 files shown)


ix0rai:
closing this for now as just about everything is conflicting -- it'll probably need a full redo if you decide to come back to it

SpaceWalkerRS:
yeah that's fair - if I do come back to it I should be targeting the 2.0 branch?

ix0rai:
2.0 is probably getting merged in the next couple weeks, you'll just want to target whatever the latest dev branch is

PR #71, opened 2023-03-18
fix inner classes displaying far too many class names

ix0rai:
this fixes that obnoxious bug where things would show:
bug
now that looks like this:
fixed :)

### DIFF
--- a/enigma/src/main/java/cuchaz/enigma/source/SourceRemapper.java
+++ b/enigma/src/main/java/cuchaz/enigma/source/SourceRemapper.java
@@ -22,12 +22,6 @@ public class SourceRemapper {
 
 			String remappedName = remapper.remap(token, movedToken);
 
-			// fixes a bug where the remapped name was displaying each and every inner class (e.g A.A.B.A.B.C instead of simply A.B.C)
-			// we do this by removing everything but the last class in the name (eg A.B.C -> C)
-			if (remappedName != null && remappedName.contains(".")) {
-				remappedName = remappedName.substring(remappedName.lastIndexOf('.') + 1);
-			}
-
 			if (remappedName != null) {
 				accumulatedOffset += movedToken.getRenameOffset(remappedName);
 				movedToken.rename(remappedSource, remappedName);
--- a/enigma/src/main/java/cuchaz/enigma/source/SourceRemapper.java
+++ b/enigma/src/main/java/cuchaz/enigma/source/SourceRemapper.java
@@ -13,7 +13,7 @@ public class SourceRemapper {
 	}
 
 	public Result remap(Remapper remapper) {
-		StringBuffer remappedSource = new StringBuffer(this.source);
+		StringBuilder remappedSource = new StringBuilder(this.source);
 		Map<Token, Token> remappedTokens = new HashMap<>();
 
 		int accumulatedOffset = 0;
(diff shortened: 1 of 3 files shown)


enigma/src/main/java/cuchaz/enigma/source/SourceRemapper.java
```diff
@@ -13,14 +13,21 @@ public SourceRemapper(String source, Iterable<Token> tokens) {
...
 
 		int accumulatedOffset = 0;
 		for (Token token : this.tokens) {
 			Token movedToken = token.move(accumulatedOffset);
 
 			String remappedName = remapper.remap(token, movedToken);
+
+			// fixes a bug where the remapped name was displaying each and every inner class (e.g A.A.B.A.B.C instead of simply A.B.C)
+			// we do this by removing everything but the last class in the name (eg A.B.C -> C)
+			if (remappedName != null && remappedName.contains(".")) {
```
SpaceWalkerRS, line 27: ngl this feels a bit jank, should it not be fixed in the `Remapper` implementation?

ByMartrixx:
I think this should be more decompiler specific. CFR handles fully qualified class references differently from QF

ix0rai:
ok, I'll see if I can fix this in a better place

ix0rai:
hmm, thing is if I put it somewhere like where quiltflower decompiles the source it doesn't do anything
the issue is uniquely present on entries that are renamed

ix0rai:
yeah after spending a while looking I don't think there's a good place to put this

SpaceWalkerRS:
One way to fix this is to have `getSourceRemapName` always return the class' simple name. It's not quite ideal, as it can make class references ambiguous (is this an inner class or a top level class that happens to have the same name?) but on the other hand it'd be more consistent overall as un-mapped class references already seem to do this.

ix0rai:
I'm happy with that fix, I'll go with that
I don't think the ambiguity issue is much of a concern since there are still the outer class tokens for you to read
thank you!

enigma/src/main/java/cuchaz/enigma/translation/representation/entry/ClassEntry.java
```diff
@@ -238,15 +238,6 @@ public static String getInnerName(String name) {
 		return name;
 	}
 
-	@Override
-	public String getSourceRemapName() {
```
SpaceWalkerRS, line 242: Correct me if I'm wrong, but the default implementation of `getSourceRemapName` returns `this.getName()` right? And `getName` returns the fully qualified name for non-inner classes.

ix0rai: oh yes that's true, I was looking at the wrong thing when I determined it could be entirely removed

SpaceWalkerRS (approved):

ByMartrixx (approved):
I'd rather have a QF specific solution, but if this works, it works

SpaceWalkerRS:
Perhaps an idea for a feature in QF is a setting for how it dumps tokens for inner classes? Because it's the only decompiler Enigma uses that dumps tokens for all the enclosing classes as well as the inner class for each reference to an inner class. The other decompilers only dump a token for the inner class.

PR #84, opened 2023-04-02
Optimize NestedPackages by deferring sorting

BasiqueEvangelist:
`NestedPackages` uses a linear search to add in new entries right now. Unsurprisingly, this is really bad when there are a lot of entries.
This PR changes the tree node type to use a non-synchronized tree node class that defers sorting, which makes adding nodes a lot faster.


ix0rai (approved):
this is brilliant! I have two little code style nitpicks: please reference fields with `this.` and put braces around all if statements. thank you!

ix0rai:
oh one other thing: `SortedMutableTreeNode` would fit better in the `node` package than the `util` package

Issue #29, opened 2023-01-17
[feature request] improved search


ix0rai:
Enigma's search is pretty good currently, but it falls short in two areas:
1. convenience: if you want to search for something, you have to know whether you're searching for a method, field or class first. This is bad for two reasons: 1, if you only remember the name of what you're looking for, and 2, if you want to search for all instances of a name, disregarding type.
2. coverage: enigma's search currently has no way to search parameters, which is a major issue when refactoring.

so, solutions!
- [ ] implement parameter searching
- [x] new "all" search that searches all four types (parameters, classes, methods, and fields)

concerns:
indexing parameters will add a lot of indexing time, so ideally that can run on a separate thread after everything else is indexed, to allow usage earlier.

notes:
should also fix https://github.com/FabricMC/Enigma/issues/408 while we're at it

PR #93, opened 2023-04-12
allow searching multiple entry types simultaneously

ix0rai:
- makes the default search keybind (shift + c) search all classes
- adds a little bar of checkboxes above search results that lets you select what to search
- probably needs a bit more work, just getting this open a little early

works towards #29


enigma-swing/src/main/java/cuchaz/enigma/gui/search/SearchUtil.java
```diff
@@ -134,7 +136,10 @@ public float getScore(String term, int hits) {
 			float maxScore = (float) Arrays.stream(this.components)
 					.mapToDouble(name -> getScoreFor(ucTerm, name))
 					.max().orElse(0.0);
-			return maxScore * (hits + 1);
+
+			// modify by type
+			int typeModifier = ((Math.abs(this.searchEntry.getType().ordinal() - SearchDialog.Type.values().length)) + 1);
```
ByMartrixx, line 141: The `abs` isn't needed since `ordinal()` is always less than `values().length`. If the goal of this is to put some entries before others I don't think the `+ 1` is needed either
```suggestion
			int typeModifier = SearchDialog.Type.values().length - this.searchEntry.getType().ordinal() + 1;
```

enigma-swing/src/main/java/cuchaz/enigma/gui/elements/MenuBar.java
```diff
@@ -376,9 +386,9 @@ private void onFontClicked(Gui gui) {
 		FontDialog.display(gui.getFrame());
 	}
 
-	private void onSearchClicked(SearchDialog.Type type) {
+	private void onSearchClicked(SearchDialog.Type[] types, boolean clear) {
```
ByMartrixx, line 389: These parameters could be reordered to use varargs instead

ByMartrixx (approved):

enigma-swing/src/main/java/cuchaz/enigma/gui/search/SearchUtil.java
ix0rai: the +1 was a safety check to make sure the modifier never resolved to zero

PR #98, opened 2023-05-13
fix some docker-related issues

ix0rai:
- no more stupid structure docker opening when you open a class
- preserve docker size when they are being opened
- fix right dock opening too small

### DIFF
--- a/enigma-swing/src/main/java/cuchaz/enigma/gui/Gui.java
+++ b/enigma-swing/src/main/java/cuchaz/enigma/gui/Gui.java
@@ -407,15 +407,6 @@ public class Gui {
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


ByMartrixx (approved):
Rip structure docker opening alongside class :pensive:

ByMartrixx:
Wait, is the docker updated when changing classes with it open?

ByMartrixx (changes requested):
Nope, it isn't updated, please fix - calling `Gui#updateStructure` where `#showStructure` was called before should be enough, I think

ix0rai:
whoops

ByMartrixx (approved):

PR #102, opened 2023-05-25
Show inner classes in bytecode view

ByMartrixx:


enigma/src/main/java/cuchaz/enigma/source/bytecode/BytecodeSource.java
```diff
@@ -10,13 +10,16 @@
...
 import java.io.StringWriter;
+import java.util.List;
 
 public class BytecodeSource implements Source {
 	private final ClassNode classNode;
+	private final List<ClassNode> otherClassNodes;
 	private final EntryRemapper remapper;
 
-	public BytecodeSource(ClassNode classNode, EntryRemapper remapper) {
+	public BytecodeSource(ClassNode classNode, List<ClassNode> otherClassNodes, EntryRemapper remapper) {
```
ix0rai, line 20: maybe refactor this so it's a clearer why there's a list of "other class nodes"? just looking at this file it would be hard to understand

ix0rai (commented):
lovely, I just think the code could be a tad clearer!

ix0rai (approved):

PR #105, opened 2023-05-30
implement crash history

ix0rai:
- also delete ExceptionIgnorer, because hey that's something we can do
- crash history is shown as a menu in the file menu where you can click on each crash to reopen the dialog


enigma-swing/src/main/java/cuchaz/enigma/gui/elements/MenuBar.java
```diff
@@ -447,6 +452,8 @@ public void onStartServerClicked() {
 			JOptionPane.showMessageDialog(this.gui.getFrame(), e.toString(), I18n.translate("menu.collab.server.start.error"), JOptionPane.ERROR_MESSAGE);
 			this.gui.getController().disconnectIfConnected(null);
 		}
+
+		throw new RuntimeException();
```
ByMartrixx, line 456: What does this do?

ByMartrixx (approved):

enigma-swing/src/main/java/cuchaz/enigma/gui/elements/MenuBar.java
ix0rai: oh oops I was using that to test
will fix once I get home

PR #114, opened 2023-06-10
fix enum statistics counting

### LINKED ISSUE #113: issues generating stats for enums
- enums always have 1 field that is always marked as mapped, even on untouched classes
- enums always seem to have extra parameters that are not mappable

orifu:
fixes this by
- no longer counting synthetic parameters in the stats (unless specified)
- marking the parameter in the `valueOf` method of enums as unrenamable

fixes #113


ix0rai (approved):
this looks perfect, thank you so much!

PR #126, opened 2023-07-16
automatic mapping format detection

ix0rai:
- standardises the way enigma determines the mapping format of any given file
- adds a test to make sure that works!

notes:
- tiny v1 currently has no contents because I don't know where to get a tiny v1 file (maybe ancient yarn?)
- maybe we should change the enigma-swing "open mappings" menu to use this?


enigma-cli/src/main/java/cuchaz/enigma/command/Command.java
```diff
@@ -99,49 +98,14 @@ public static EnigmaProject openProject(Path fileJarIn, Path fileMappings, Enigm
...
-			}
-		}
-
-		RuntimeException exception = new RuntimeException("Unable to write mappings!");
-		for (Exception supressedException : suppressed) {
-			exception.addSuppressed(supressedException);
-		}
-
-		throw exception;
+	protected static void writeMappings(EntryTree<EntryMapping> mappings, Path path, ProgressListener progress, MappingSaveParameters saveParameters) {
```
ByMartrixx, line 106: I think it would be better to entirely remove this method, here `path` might not be an existing file, and the only command that uses it is actually meant to use the same format as its input.

enigma-cli/src/main/java/cuchaz/enigma/command/Command.java
```diff
@@ -253,7 +217,7 @@ protected static <T> void writeDebugDelta(DeltaTrackingTree<T> tree, Path output
 	}
 
 	public static class ConsoleProgressListener implements ProgressListener {
-		private static final int ReportTime = 5000; // 5s
```
ByMartrixx, line 256: :vomiting_face:

enigma/src/main/java/cuchaz/enigma/translation/mapping/serde/MappingFormat.java
```diff
@@ -67,25 +69,60 @@ public MappingsReader getReader() {
...
+					try {
+						String contents = Files.readString(file);
+						String firstLine = contents.split("\n")[0];
+						String[] splitFirstLine = firstLine.split(" ");
+
+						if (splitFirstLine.length == 3 && splitFirstLine[1].equals("->") && splitFirstLine[2].endsWith(":")) {
+							return PROGUARD;
+						}
+					} catch (IOException e) {
+						return RECAF;
```
ByMartrixx, line 120: Isn't this `IOException` only possible if the file can't be read?

ix0rai: yes, I think that was a bad decision on my part
should this just rethrow?

ByMartrixx: Yeah, a RuntimeException should do

PR #129, opened 2023-07-25
Migrate to Kotlin DSL

Decencies:
Migrate the groovy buildscripts to the Kotlin DSL.


kb-1000:
It's worth noting Kotlin DSL has an overload of the `maven` method taking an `url` parameter, so it's possible to write them like
```kotlin
maven(url = "https://maven.quiltmc.org/repository/release/")

**Comment by Decencies** [comment]
Any movement on this? Can this be merged?

**Comment by ix0rai** [comment]
@ByMartrixx do you want to take a look first? I poked through and everything looks good. I wouldn't mind slipping it in without your review but I want to make sure

**Review by ByMartrixx — APPROVED** [review]

**Review by EnnuiL — COMMENTED** [review]
Hm, about this PR: I'm not really sure about it; this feels more like personal preference than something that would benefit the project;
like, despite build.gradle.kts being nicer and such? it does end torpedoing support with any IDEs that aren't IntelliJ; hence why the only Quilt projects that do have it are like, small projects that people won't really use directly nor touch with frequency (ex: the Quilt Legacy Installer, which is supposed to be dead already in a good timeline!)

those are only my thoughts though, i don't know about the others' opinions about this

**Comment by SpaceWalkerRS** [comment]
As an Eclipse user I'm not too fond of this change.

**Comment by TheKodeToad** [comment]
It should work with Eclipse.
there is a kotlin plugin but it doesn't work too well :/
however, import works fine (afaik) :D

**Comment by ix0rai** [comment]
ok, I've looked into this myself and my Official Decision is:
Since `.kts` support in Eclipse is limited to simply It Works (even after installing every search result for Kotlin (not kidding), there was still not even syntax highlighting in the buildscript), I don't think we can go through with this -- yet. I still agree with this change though, so once Eclipse gets its act together in terms of supporting Kotlin scripting with Gradle we can merge this. We can't just say "develop Enigma in Intellij or have fun looking at a text file for a buildscript".

**Comment by EnnuiL** [comment]
> ok, I've looked into this myself and my Official Decision is: Since `.kts` support in Eclipse is limited to simply It Works (even after installing every search result for Kotlin (not kidding), there was still not even syntax highlighting in the buildscript), I don't think we can go through with this -- yet. I still agree with this change though, so once Eclipse gets its act together in terms of supporting Kotlin scripting with Gradle we can merge this. We can't just say "develop Enigma in Intellij or have fun looking at a text file for a buildscript".

While I do wanna be as optimistic for IDE support on other IDEs like Eclipse and VSCode? i really doubt that it's something that will be accomplished any time soon; which is a shame, because kotlin script can be nicer indeed

**Comment by Decencies** [comment]
> > ok, I've looked into this myself and my Official Decision is: Since `.kts` support in Eclipse is limited to simply It Works (even after installing every search result for Kotlin (not kidding), there was still not even syntax highlighting in the buildscript), I don't think we can go through with this -- yet. I still agree with this change though, so once Eclipse gets its act together in terms of supporting Kotlin scripting with Gradle we can merge this. We can't just say "develop Enigma in Intellij or have fun looking at a text file for a buildscript".
> 
> 
> 
> While I do wanna be as optimistic for IDE support on other IDEs like Eclipse and VSCode? i really doubt that it's something that will be accomplished any time soon; which is a shame, because kotlin script can be nicer indeed

Unfortunately it's up to the IDE to provide the best end user experience, not the toolchain itself. While it would be nice to be more inclusive and support multiple IDE configurations, I think using IntelliJ is basically the standard in 2023, especially for Kotlin development, as it is the official Kotlin IDE. Whilst this project isn't bound to Kotlin, I still believe that using IDEA as the base IDE for development is certainly one of the best choices for the potential progression and expansion of the project. Perhaps the enigma enigma itself will have an IDEA frontend at some point(?)

Again, this is open to suggestions, but as pointed out already, Kotlin build scripts are marginally better than their groovy counterparts. We have one of 2 tradeoffs, very shoddy highlighting & completion in the groovy buildscript or or complete highlighting and completion support in the Kotlin script.

File: enigma-swing/build.gradle.kts [hunk 1]
```diff
@@ -0,0 +1,58 @@
...
+}
+
+fun Project.registerTestTask(name: String) {
+    tasks.register("${name}TestGui", JavaExec::class.java) {
+        group = "test"
+        dependsOn(":enigma:${name}TestObf")
+        mainClass.set(application.mainClass)
+        classpath = sourceSets["main"].runtimeClasspath
+
+        val jar = project(":enigma").file("build/test-obf/${name}.jar")
```
sschr15, line 32:
```suggestion
        val jar = project(":enigma").file("build/test-obf/$name.jar")
```

enigma/build.gradle.kts
```diff
@@ -0,0 +1,107 @@
...
+        group = "test-setup"
+        dependsOn("${name}TestJar")
+        mainClass.set("proguard.ProGuard")
+        classpath(configurations["proGuard"])
+
+        args(
+            "@$confFile",
+            "-injars", file("build/test-inputs/${name}.jar"),
+            "-libraryjars", libraryJarsArg,
+            "-outjars", file("build/test-obf/${name}.jar")
```
sschr15, line 74:
```suggestion
            "-injars", file("build/test-inputs/$name.jar"),
            "-libraryjars", libraryJarsArg,
            "-outjars", file("build/test-obf/$name.jar")
```

enigma/build.gradle.kts
```diff
@@ -0,0 +1,107 @@
...
+            dependsOn("${f.name}TestObf")
+        }
+    }
+}
+
+publishing {
+    publications {
+        create<MavenPublication>(project.name) {
+            groupId = project.group.toString()
+            artifactId = project.name.toString()
```
sschr15, line 102: Same as prior instances

ix0rai:
closing this for now -- when eclipse improves support, we can base a new buildscript rework off of these changes

Decencies:
Let's face it, Eclipse is never going to implement Kotlin support fully as they are basically implementing technologies  from "competing" editors. In my humble opinion I think closing this issue based on IDE support is nonsense. Kotlin can also be installed standalone, assuming users are already prepared to suffer greatly from using eclipse, I'm almost certain installing Kotlin as a standalone technology will be the least of their troubles.

I digress...

Also for your information, all the client and server-side improvements I have written for enigma are written in Kotlin and weren't really designed for the language constraints of Java.

PR #135, opened 2023-08-05
remove access modifier editing

ix0rai:
probably good, but I have yet to look into whether I've broken enigma's mapping format by removing all the modifier code from there!


enigma/src/main/java/cuchaz/enigma/classprovider/ObfuscationFixClassProvider.java
```diff
@@ -18,7 +18,7 @@
  * Wraps a ClassProvider to apply fixes to the following problems introduced by the obfuscator,
  * so that the classes can be decompiled correctly:
  * <ul>
- * 	<li>Bridge methods missing the "bridge" or "synthetic" access modifier
+ * 	<li>Bridge methods missing the "bridge" or "synthetic" zaccess modifier
```
IotaBread, line 21: zaccess

IotaBread (commented):
The protocol version should change, maybe we could do what Minecraft does for snapshots?

enigma/src/main/java/cuchaz/enigma/classprovider/ObfuscationFixClassProvider.java
ix0rai: HOW

ix0rai:
@IotaBread not familiar with what minecraft does for snapshots, what are you suggesting?

IotaBread:
Basically, [this javadoc in SharedConstants#getProtocolVersion](https://github.com/QuiltMC/quilt-mappings/blob/23w32a/mappings/net/minecraft/SharedConstants.mapping#L118-L119), where `SNAPSHOT_PROTOCOL_VERSION_EXTRA_BITS` is a constant (30 iirc), and `SNAPSHOT_PROTOCOL_VERSION` increases every snapshot

ix0rai:
that's quite clever actually
I'm not going to throw that into the pr, but it would be nice to have

PR #137, opened 2023-08-11
Ensure Tiny V2 always writes in the same order

IotaBread:
Non-deterministic output is quite annoying when comparing two mapping files with very few actual differences


ix0rai (approved):
brilliant, love the tests!

PR #138, opened 2023-08-11
remove tiny v1 and improve commands to use automatic format detection

ix0rai:
- [x] STILL NEED TO FIX ARGS

fairly simple pr despite the huge diff, just need to make sure I've gotten all references


enigma-cli/src/main/java/cuchaz/enigma/command/MappingCommandsUtil.java
```diff
@@ -59,33 +22,16 @@ public static void write(EntryTree<EntryMapping> mappings, String type, Path pat
...
-			format = MappingFormat.valueOf(type.toUpperCase());
+			format = MappingFormat.valueOf(type);
 		} catch (IllegalArgumentException ignored) {
-			// ignored!
-		}
-
-		if (format != null) {
-			format.getWriter().write(mappings, path, ProgressListener.none(), saveParameters);
-			return;
+			format = MappingFormat.valueOf(type.toLowerCase());
```
IotaBread, line 32:
```suggestion
			format = MappingFormat.valueOf(type.toUpperCase());
```
All of `MappingFormat`'s fields are upper case

IotaBread (approved):

PR #143, opened 2023-08-26
Fix regex for renaming package with a capital letter

uniformization:
The current input check does not account for capitalization in the package path. It will fail whenever it does.
It's possible packages could have capital letters so I'm making this PR to fix that.

Before:
M6gEzZEBtC

After:
uGcete0vrj


ix0rai (approved):
conscious decision (I have strong opinions on code style), but yeah this should probably be allowed

PR #151, opened 2023-09-13
add "move package" option to context menu

ix0rai:
turns out package renaming can cause unexpected behaviour, so this is my fix. "move package" will only refactor the top-level package, to match expectations

other changes:
- fix cancel button in package name gui
- more docs
- add a test for moving package


IotaBread (approved):
Looks great!

PR #152, opened 2023-09-27
stats rework

ix0rai:
- adds stats to packages
- improves speed for individual class generation via discovering entries recursively for each class
- will decrease speed for full-project stats gen but it's a tradeoff to make the more common stats usage work better
- stats dialog opens instantly
- stats tree opens instantly
- stats for the full project are generated at startup, so it takes longer to populate but is overall a better experience

todo
- [x] reimplement progress listener
- [x] testing
- [x] documentation
- [x] pretty sure that bug where it shows as unmapped during gen is back

notes:
- [x] should check if the recursive discovery logic is totally duplicated from MappingValidator


enigma-swing/src/main/java/cuchaz/enigma/gui/GuiController.java
```diff
@@ -539,17 +546,19 @@ private void applyChange0(ValidationContext vc, EntryChange<?> change, boolean u
 		}
 	}
 
-	public void openStats(Set<StatType> includedTypes, String topLevelPackage, boolean includeSynthetic) {
+	public void openStatsTree(Set<StatType> includedTypes) {
 		ProgressDialog.runOffThread(this.gui, progress -> {
-			String data = this.gui.getStatsManager().getGenerator().generate(progress, includedTypes, topLevelPackage, includeSynthetic).getTreeJson();
+			StatsResult overall = this.getStatsGenerator().getResultNullable().getOverall();
+			StatsTree<Integer> tree = overall.buildTree(UiConfig.getLastTopLevelPackage(), includedTypes);
+			String treeJson = new GsonBuilder().setPrettyPrinting().create().toJson(tree.root);
```
IotaBread, line 553: I think this `Gson` instance should be static

IotaBread (approved):
Great work

PR #154, opened 2023-10-05
Api/impl separation

### LINKED ISSUE #109: [feature request] add a clear api/impl separation in `enigma` and `enigma-server`
as a step towards decoupling `enigma` from `enigma-swing` and making it easier to create more clients, `enigma` should clearly establish what is meant to be used and what is not. this will also allow users to follow semver with enigma, as it makes it easy for us to tell what is public api and should bump major when changed.

IotaBread:
Closes #109. For now I only moved classes that are clearly one of api or impl, I have to figure out how to classify the rest


ix0rai (commented):
looks great! only thing I'm wondering about is if we should make `Decompilers` API, it doesn't really make sense to lock it up when `DecompilerService` is public

ix0rai (approved):

PR #155, opened 2023-10-05
toss everything deprecated

ix0rai:
removes the last of the deprecated things. one thing I'm not sure about removing is the `getSource` overload, I don't see much of a point to removing it


enigma/src/main/java/org/quiltmc/enigma/source/Decompiler.java
```diff
@@ -4,10 +4,5 @@
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 public interface Decompiler {
-	@Deprecated // use remapper specific one for easy doc inclusion
-	default Source getSource(String className) {
```
IotaBread, line 8: Id keep this method, or rename it to `getUndocumentedSource`

ix0rai: that's what I was thinking!

enigma/src/main/java/org/quiltmc/enigma/source/Decompiler.java
```diff
@@ -4,10 +4,9 @@
...
-		return this.getSource(className, null);
+	default Source getUndocumentedSource(String className) {
+		return this.getUndocumentedSource(className, null);
 	}
 
-	Source getSource(String className, @Nullable EntryRemapper remapper);
+	Source getUndocumentedSource(String className, @Nullable EntryRemapper remapper);
```
IotaBread, line 11: This one does produce documented source :p

IotaBread: At least if remapper isn't null

ix0rai: throwing up how did I not notice that happened :p

ix0rai: ah intellij overload renaming
fixed!

IotaBread (approved):

Issue #20, opened 2022-11-30
[feature request] package creation confirmation


ix0rai:
currently, enigma automatically creates packages according to the names you give your files.

this is fine, but gets annoying if you have a typo in your package name and put your file in the wrong place. A confirmation dialogue with "confirm" and "cancel" options would be a simple solution to this issue.

ix0rai:
hm, one thing blocking this is that to check that the package doesn't exist you would have to run through all classes, which would take ages
maybe an index of packages could solve that?

PR #156, opened 2023-10-09
add mappings indexing

### LINKED ISSUE #111: [feature request] mappings indexing
implement a system to index the mappings file that works similarly to jar indexing. this will be useful for many things, with two indexes already planned:
- an index of javadoc links, to implement https://github.com/OrnitheMC/enigma/issues/27
- an index of package names, to implement https://github.com/QuiltMC/enigma/issues/20

ix0rai:
fixes #111 
I'm going to add a package name index for fixing #20 before this is merged as a proof of concept


ix0rai:
still todo:
figure out a solution for reindexing mappings when they're changed, without running over the entire thing again
(maybe save maps of entry -> index value for everything?)

ix0rai:
pr done!

enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/mapping/MappingsIndex.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.Set;
+
+public class MappingsIndex implements MappingsIndexer {
+	private final Collection<MappingsIndexer> indexers;
+	private final PackageIndex packageIndex;
+
+	private ProgressListener progress;
+	private int work;
+
+	public MappingsIndex(PackageIndex packageIndex) {
```
IotaBread, line 27: I feel like this could be implemented as a `Map<Class<? extends MappingsIndexer>, MappingsIndexer>` for indexers collection, and here `MappingsIndexer... indexers` as a parameter, so you keep any arbitrary mapping indexers instead of having a field for each one of them... though I'm not sure if it'd useful at all...

IotaBread (approved):

enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/mapping/MappingsIndex.java
ix0rai: true, I just copied the way `JarIndex` does it. I'll change things up tomorrow!

IotaBread (approved):

PR #158, opened 2023-11-02
Allow using json lists for service args

IotaBread:
Instead of typing a list as a comma separated string, you can write using json syntax now
```json
{
  "services": {
  "jar_indexer": {
    "id": "enigma:foo",
    "args": {
      "abc": [
        "foo",
        "bar",
        "baz"
      ]
    }
  }
}
```

Would fix the extra long string in the QM profile:
https://github.com/QuiltMC/quilt-mappings/blob/cb88c6627f36721f036e6a1370c14e356cedcdda/enigma_profile.json#L14


ix0rai:
the `Either` class is great util -- I'm all for including it in enigma -- but I think that we shouldn't use it to preserve the old syntax. it's much more fragile (IDEs will recognise JSON errors, but not you forgetting a comma in a string) and 2.0 is our chance to get rid of weird old stuff

ix0rai (approved):
nvm misunderstood!

PR #167, opened 2023-11-30
move to quilt config

ix0rai:
new config is organised into five files: `net`, `decompiler`, `keybind`, `main`, and `docker`. config structure is pretty similar to the old one. it's subjective, but I like the way you interact with the new config much better than the old one.


ix0rai:
before this is merged I want to integrate a TOML serialiser into qconf, as well as a SerializedName annotation

ix0rai:
pr is just about done. note that until https://github.com/QuiltMC/quilt-config/pull/14 gets merged, you'll have to clone https://github.com/ix0rai/quilt-config/tree/master and publish it to maven local in order to test

enigma-swing/src/main/java/org/quiltmc/enigma/gui/config/Config.java
```diff
@@ -0,0 +1,212 @@
...
+import org.quiltmc.enigma.gui.dialog.EnigmaQuickFindDialog;
+import org.quiltmc.enigma.util.I18n;
+import org.quiltmc.syntaxpain.SyntaxpainConfiguration;
+
+import java.awt.Dimension;
+import java.awt.Point;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public final class Config extends ReflectiveConfig {
```
ix0rai, line 25: should provide some docs on config structure here

IotaBread (approved):

PR #168, opened 2023-12-07
2.0 API and impl changes for name proposal

IotaBread:
Still a work in progress. `MergedEntryMappingTree` isn't finished yet, but its meant to replace the `mergedNames` tree in `EntryRemapper` so both of its underlying trees are updated accordingly. It will need testing to ensure the tree works as intended and doesn't break other features


enigma/src/main/java/org/quiltmc/enigma/api/translation/representation/entry/Entry.java
```diff
@@ -158,6 +158,15 @@ default ClassEntry getTopLevelClass() {
 		return Objects.requireNonNull(last, () -> String.format("%s has no top level class?", this));
 	}
 
+	/**
+	 * Get the complete ancestry list of this entry, including itself.
+	 * An entry is considered an ancestor if it's the parent of the current entry, or another ancestor of it.
```
ix0rai, line 163:
```suggestion
	 * Searches recursively: an entry is considered an ancestor if it's the parent of the current entry or any of its parents.
```

enigma/src/main/java/org/quiltmc/enigma/api/translation/representation/entry/Entry.java
```diff
@@ -169,6 +178,15 @@ default List<Entry<?>> getAncestry() {
 		return entries;
 	}
 
+	/**
+	 * Find the closest ancestor of the given entry type.
+	 *
+	 * @param <E> the entry type to search for
+	 * @param type the class of the entry type to search for.
```
ix0rai, line 185:
```suggestion
	 * @param type the class of the entry type to search for
```

enigma/src/main/java/org/quiltmc/enigma/api/translation/representation/entry/Entry.java
```diff
@@ -183,6 +201,18 @@ default <E extends Entry<?>> E findAncestor(Class<E> type) {
 		return null;
 	}
 
+	/**
+	 * Replace an entry in the ancestry of this one.
```
ix0rai, line 205:
```suggestion
	 * Replaces an entry in the ancestry of this entry.
```

enigma/src/main/java/org/quiltmc/enigma/api/translation/representation/entry/MethodEntry.java
```diff
@@ -49,29 +50,45 @@ public boolean isConstructor() {
 	}
 
 	/**
-	 * Creates an iterator of all parameters in this method. Unmapped args will have no name, and javadoc is ignored.
+	 * Get a list of all parameters in this method.
```
ix0rai, line 53:
```suggestion
	 * Get a list of all parameters in this method. All parameters will have an empty name.
```
I think the clarification is important (read: I will forget and wonder why my code doesn't work)

enigma/src/main/java/org/quiltmc/enigma/api/translation/representation/entry/MethodEntry.java
```diff
@@ -49,29 +50,45 @@ public boolean isConstructor() {
...
-				argIndex += arg.getSize();
+				i += arg.getSize();
 			}
 		}
 
+		return parameters;
+	}
+
+	/**
+	 * Creates an iterator of all parameters in this method, also doing translation. Unmapped args will have no name, and javadoc is ignored.
```
ix0rai, line 77:
```suggestion
	 * Creates an iterator of all parameters in this method, also doing translation. Unmapped args will have an empty name, and javadoc is ignored.
```
now that I've changed the other one I think this is better because it tells you it's an empty string instead of leaving you wondering if it's null or empty

enigma/src/main/java/org/quiltmc/enigma/api/translation/representation/entry/Entry.java
```diff
@@ -183,6 +201,18 @@ default <E extends Entry<?>> E findAncestor(Class<E> type) {
...
+	 * If the target entry is this one, returns the replacement instead.
+	 * If the target and replacement entries are the same, returns this entry without changes.
+	 *
+	 * @param <E> the type of the entry to replace
+	 * @param target the entry to replace
+	 * @param replacement the replacement entry
+	 * @return an entry with the same ancestry as this one, with the replacement applied.
```
ix0rai, line 212:
```suggestion
	 * @return an entry with the same ancestry as this one, with the replacement applied
```

ix0rai (commented):
fantastic! just javadoc nitpicks

PR #169, opened 2023-12-21
Keep JAR_PROPOSED mappings separate from other proposals

IotaBread:
Now all JAR_PROPOSED mappings are restored when reloading mappings. If they were overridden by a dynamic proposal before, they wouldn't be recovered


ix0rai (approved):
much better solution!

PR #171, opened 2023-12-23
Tokenize the ASM bytecode view

### LINKED ISSUE #40: [feature request] Tokenization / remapping in ASM view
Currently the Bytecode mode does not contain any tokenization or remapping capabilities, preventing any mapping from being done while using that "decompiler" mode. In some cases, being able to map based on disassembly would be beneficial.

IotaBread:
Fixes #40. Only thing left is to display the constructors of mapped classes as `<init>`s

image

### DIFF
--- a/enigma/src/main/java/org/quiltmc/enigma/impl/source/bytecode/EnigmaTextifier.java
+++ b/enigma/src/main/java/org/quiltmc/enigma/impl/source/bytecode/EnigmaTextifier.java
@@ -186,7 +186,7 @@ public class EnigmaTextifier extends Textifier {
 		this.queueToken(new QueuedToken.Array(
 			new QueuedToken.Descriptor(descriptor), // field descriptor
 			new QueuedToken.OffsetToken(descriptor.length() + 1, name,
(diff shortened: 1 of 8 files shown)


enigma/src/main/java/org/quiltmc/enigma/impl/source/bytecode/EnigmaTextifier.java
```diff
@@ -1,18 +1,522 @@
...
+			public boolean shouldSkip() {
+				return !this.descriptor.contains("L");
+			}
+		}
+
+		record MethodDescriptor(String descriptor) implements QueuedToken {
+		}
+
+		record Signature() implements QueuedToken {
+		} // TODO
```
ix0rai, line 48: stray todo?

ix0rai (approved):
looks amazing!

ix0rai:
note that this should be merged into either master or a develop/2.1, not dev/2.0

enigma/src/main/java/org/quiltmc/enigma/impl/source/bytecode/EnigmaTextifier.java
IotaBread: Oh yeah, let me remove it

PR #172, opened 2023-12-29
fixes to mappings index

ix0rai:
- mappings index was skipping entries due to me using my own stupid way of collecting nodes
- package index only indexed the last package in the heirarchy for each class
- package index depended on one sole class for each package, meaning if you renamed that one it "forgot" that package existed
- the message was sometimes a little bit misleading, since unmapped classes don't get their packages indexed

### DIFF
--- a/enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/mapping/PackageIndex.java
+++ b/enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/mapping/PackageIndex.java
@@ -6,7 +6,6 @@ import org.quiltmc.enigma.api.translation.mapping.EntryMapping;
 import org.quiltmc.enigma.api.translation.representation.entry.ClassEntry;
 import org.quiltmc.enigma.api.translation.representation.entry.Entry;
 
-import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/mapping/PackageIndex.java
```diff
@@ -1,35 +1,51 @@
...
+			// index all different versions of the package name
+			// note we're not avoiding duplicates: otherwise reindexing would erroneously remove package names
+			List<String> names = new ArrayList<>();
+			while (name != null && name.contains("/")) {
+				name = ClassEntry.getParentPackage(name);
+				names.add(name);
+			}
+
+			for (String p : names) {
+				this.packageNames.put(entry, p);
```
IotaBread, line 40: Why not add them directly to the map instead of to `names` and then to it?

ix0rai: excellent point

IotaBread (approved):

PR #173, opened 2023-12-29
Add a couple more development options

IotaBread:


enigma-swing/src/main/java/org/quiltmc/enigma/gui/panel/IdentifierPanel.java
```diff
@@ -150,6 +151,11 @@ public void refreshReference() {
 			} else {
 				throw new IllegalStateException("unreachable");
 			}
+
+			if (Config.main().development.showMappingSourcePlugin.value()) {
+				var mapping = this.gui.getController().getProject().getRemapper().getMapping(this.entry);
+				th.addStringRow("Source plugin", mapping.sourcePluginId());
```
ix0rai, line 157: extremely nitpicky but this should be translatable

ix0rai (approved):
nice!

PR #174, opened 2024-01-01
Resolve the correct amount of entries when renaming a method

IotaBread:
If two unrelated method declarations have the same name and descriptor, you can override both of them at the same time by extending their owners. If you renamed only one of these, the other one would either reset to the default impl or break during runtime.

With this pr, renames apply to both of them at the same time.

### DIFF
--- a/enigma/src/test/java/org/quiltmc/enigma/translation/mapping/EntryRemapperTest.java
+++ b/enigma/src/test/java/org/quiltmc/enigma/translation/mapping/EntryRemapperTest.java
@@ -1,0 +1,110 @@
+package org.quiltmc.enigma.translation.mapping;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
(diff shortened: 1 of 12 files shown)


enigma/src/test/java/org/quiltmc/enigma/translation/mapping/EntryRemapperTest.java
```diff
@@ -0,0 +1,122 @@
...
+		assertName(TestEntryFactory.newMethod("d", "a", "()F"), name);
+
+		name = "unionDC";
+		remapper.putMapping(newVC(), TestEntryFactory.newMethod("d", "a", "()F"), new EntryMapping(name));
+
+		assertName(TestEntryFactory.newMethod("d", "a", "()F"), name);
+		assertName(TestEntryFactory.newMethod("c", "a", "()F"), name);
+	}
+
+	private static ValidationContext newVC() {
```
ix0rai, line 106: this stuff should probably be moved to our common testing utils since it's in multiple files now (also in `TestMappingValidator`)

ix0rai (approved):
I appreciate that you wrote up documentation for this :)

OroArmor (approved):

Issue #192, opened 2024-04-04
enigma hangs on `Done!` after completing indexing


ix0rai:
according to @NebelNidas, this scales with project size and makes quilt enigma take prohibitively long to boot up on large projects

PR #196, opened 2024-04-06
fix all those issues I just made

### LINKED ISSUE #193: allow disabling stat icons
stat icons are *very* slow. it should be possible to disable them, and perhaps we should implement a cap on top-level classes before they're automatically disabled

### LINKED ISSUE #194: always show dev menu if a development option is enabled

ix0rai:
fix #193 
fix #194 
will make attempts at #192


enigma-swing/src/main/java/org/quiltmc/enigma/gui/config/DevSection.java
```diff
@@ -1,21 +1,20 @@
 package org.quiltmc.enigma.gui.config;
 
-import com.google.gson.annotations.SerializedName;
```
IotaBread, line 3: :p

ix0rai:
@IotaBread imo we merge this unfinished for 2.3 along with the ASM fix and then address the last issue in the next update

IotaBread:
sounds good to me!

ix0rai:
gorgeous let's do it

PR #203, opened 2024-04-24
fix entry remapper resolving nonexistent linked methods

ix0rai:
cc @IotaBread to see if the broken test is a major issue


IotaBread (changes requested):
While this would fix the bug, it would reintroduce another bug that was present before this method was introduced. The "right" fix would be to filter out non-existing entries from the resolved ones (non-existing entries shouldn't be returned from `#resolveEntry`, it's a bug; I'm working on it)

ix0rai:
got it, I'll wait for your fix!

PR #205, opened 2024-04-29
validate lambda parameters for uniqueness and add more tests

### LINKED ISSUE #188: validate duplicate variable names for lambdas
lambda names can conflict with other variables within their scope. I'm actually not sure they're validated for uniqueness at all

ix0rai:
adds lambdas into the mix for parameter uniqueness validation.

known issues:
- ~~names are not checked into lambdas (renaming a parameter of a top level method into a conflict with a parameter of an enclosed method will not throw an error)~~

fixes #188


ix0rai:
> names are not checked into lambdas (renaming a parameter of a top level method into a conflict with a parameter of an enclosed method will not throw an error)
should we even do this?

IotaBread (approved):
Looks great!

PR #214, opened 2024-08-03
update default dark theme colors

supersaiyansubtlety:
Now with darker darks and lighter lights:

<details>

<summary>Before/after 1</summary>

enigma-default-dark-theme-before-after-1

</details>

<details>

<summary>Before/after 2</summary>

enigma-default-dark-theme-before-after-2

</details>

### DIFF
--- a/enigma-swing/src/main/java/org/quiltmc/enigma/gui/config/theme/Theme.java
+++ b/enigma-swing/src/main/java/org/quiltmc/enigma/gui/config/theme/Theme.java
@@ -134,30 +134,30 @@ public class Theme extends ReflectiveConfig.Section {
 
 		public void configure(boolean dark) {
 			if (dark) {
-				setIfAbsent(this.lineNumbersForeground, new SerializableColor(0xFFA4A4A3));
-				setIfAbsent(this.lineNumbersBackground, new SerializableColor(0xFF313335));
-				setIfAbsent(this.lineNumbersSelected, new SerializableColor(0xFF606366));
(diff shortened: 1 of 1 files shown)


ix0rai:
i'll be honestly i personally like the old one better
i think the contrast between the code view and the rest of the window is a bit much
would you mind moving this into its own theme?

supersaiyansubtlety:
It would look better if the dark LaF was darker...
(I'd like to make that customizable in another PR, eventually)

I could move it to another theme, but it'd require reworking how default syntaxpane colors are loaded, currently they only allow for one dark and one light set of default colors.

Before I do that I'd like to hear what @OroArmor and @woodiertexas think of this theme.
No need to ship it with Enigma if I'll be the only one using it.

ix0rai:
i've been meaning to do a proper customisation screen in swing for ages

OroArmor:
> Before I do that I'd like to hear what @OroArmor and @ woodiertexas think of this theme.
> No need to ship it with Enigma if I'll be the only one using it.

Hm I agree with rai that I don't really like the darkness of the code compared to the sidebars. If those were darker I think I would like it more

woodiertexas:
personally, the enigma UI should be as dark as the code view

supersaiyansubtlety:
Alright, sounds like we all agree it would be better if the non-syntaxpane colors (LaF colors) where also darker.

Closing this in favor of <https://github.com/supersaiyansubtlety/enigma/tree/improve-theme-handling>.

I aim to implement LaF color customization there.
I'll open a draft PR once the `ClassCastException` I asked about on discord is sorted out.

PR #218, opened 2024-08-21
Fix `StringIndexOutOfBoundsException` when parsing recent files

PiTheGuy:


ix0rai:
can you share the config file you used to reproduce this?

PiTheGuy:
Where do I find that?

PiTheGuy:
I don't know what file you want so here's the whole folder
[config.zip](https://github.com/user-attachments/files/16698490/config.zip)

ix0rai:
thanks!

PR #216, opened 2024-08-11
Improve theme handling, add Darcerula

supersaiyansubtlety:
## Comments aren't currently added by the `Config.Creator`s I've added and I don't know why
https://github.com/QuiltMC/enigma/blob/9fed08f0b0ad2d2d827bbb31936ac36926f2a084/enigma-swing/src/main/java/org/quiltmc/enigma/gui/config/theme/properties/NonConfigurableLafThemeProperties.java#L14

https://github.com/QuiltMC/enigma/blob/9fed08f0b0ad2d2d827bbb31936ac36926f2a084/enigma-swing/src/main/java/org/quiltmc/enigma/gui/config/theme/properties/ThemeProperties.java#L40

## Renames/Moves
- `Theme.Colors` -> `SyntaxPaneProperties.Colors`
- `LookAndFeel` -> `Config.ThemeChoice`
- `Config.lookAndFeel` -> `theme`
- `Themes`-> `ThemeUtil`

## Puts theme configs in separate files under theme/
to de-clutter main.toml

## Adds `LookAndFeelProperties`
~~Currently these are present but ignored in `metalTheme`, `systemTheme`, and `noneTheme`.~~
Only themes that support `look_and_feel_colors` configuration have the corresponding properties.
Themes that don't support them have comments saying they can't be configured.

`metalTheme`, `systemTheme`, and `noneTheme` don't support look and feel configuration.

`FlatLaf`-based `LookAndFeels` are configurable using only the seven colors in `LookAndFeelColors` thanks to `FlatPropertiesLaf` which allows making `FlatLaf`s based on other `FlatLaf` implementations, which derive many color variations from a few base colors.

`MetalLookAndFeel` could be made similarly configurable, but it doesn't have built-in color deriving. I've laid out what it would look like on a separate branch in `ConfigurableMetalOceanLookAndFeel`, but it's incomplete, as explained in the commit message: 76ab7444ab16c40ceda29b8ab1d1ac714796af42.
If it's alright to bundle a color library and have a giant map of all `MetalLookAndFeel`'s keys, I can try to finish `ConfigurableMetalOceanLookAndFeel` in a followup PR.

`PrintColorKeyGroupsMapCodeTask` helps figure out what colors are actually used by a `LookAndFeel` and creates maps for keys; it created the starting point for `ConfigurableMetalOceanLookAndFeel`. I can remove it, but it  seems like it could be handy for anyone else working on `LookAndFeel`s.

## Attaches themes to `ThemeChoice`s

[...]

PR #222, opened 2024-09-13
Fix (most) scaling

supersaiyansubtlety:
### Removes `ScaleUtil::rescaleFontInConfig`
It scaled fonts in addition to `ScaleUtil::scaleFont`, and it didn't account for switching themes. If you increased scale on one theme, its fonts would be scaled to match. But if you switched themes, the previous theme would keep its scaled fonts and the new theme would not have scaled themes. If you then decreased scale on the new theme, it fonts would be scaled down (and they'd never been scaled up because the scale increase happened on another theme).

This makes the font sizes in theme configs 'base' sizes before main.toml's scale is applied.

### Fixes Theme's editor font size getting overwritten (and being read twice)
It was read and set first in `Config::updateSyntaxpain`.
Then it was read again in `ScaleUtil::applyScaling` and its size was explicitly set to 12 and scaled with main.toml's scale.

Now its read in `Config::updateSyntaxpain`, and `ScaleUtil::applyScaling` gets the syntaxpain's current font (already set from config), gets its size, and scales it with main.toml's scale.

### Fixes `DockerButton` scaled size+position for flatlaf themes
Broken since #216 (oops).
Buttons were already manually scaled+positioned to account for text width, I just added the same thing for text height.

### Fixes icon scaling for flatlaf themes
Also broken since #216 (oops).
Instead of skipping all scaling for flatlaf themes in `ScaleUtil::applyScaling`, they now have their fonts scaled.

### DIFF
--- a/enigma-swing/src/main/java/org/quiltmc/enigma/gui/config/theme/Theme.java
+++ b/enigma-swing/src/main/java/org/quiltmc/enigma/gui/config/theme/Theme.java
@@ -41,8 +41,8 @@ public class Theme extends ReflectiveConfig {
 		this.properties.configure();
 	}
 
-	public boolean needsScaling() {
-		return this.properties.needsScaling();
+	public boolean onlyScaleFonts() {
+		return this.properties.onlyScaleFonts();
 	}
 
 	public static class Fonts extends Section {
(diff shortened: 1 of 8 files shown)


ix0rai (approved):
looks good! (well, as good as code can look when it's dealing with this disaster)
i'll test this tonight!

ix0rai:
interface scales flawlessly on my laptop, thank you so much!

PR #233, opened 2024-10-15
better logging and smarter behaviour for drop invalid mappings

ix0rai:
- add a test for drop invalid mappings
- add better logging for drop invalid mappings
- improved dropping behaviour
  - drop methods that have no mapping and only invalid parameters

TODO:
- [x] fix out-of-bounds parameters not being dropped

### DIFF
--- a/enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/jar/EntryIndex.java
+++ b/enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/jar/EntryIndex.java
@@ -1,5 +1,7 @@
 package org.quiltmc.enigma.api.analysis.index.jar;
 
+import org.objectweb.asm.tree.ClassNode;
+import org.quiltmc.enigma.api.EnigmaProject;
 import org.quiltmc.enigma.api.translation.mapping.EntryMapping;
 import org.quiltmc.enigma.api.translation.mapping.tree.EntryTree;
 import org.quiltmc.enigma.api.translation.mapping.tree.HashEntryTree;
@@ -17,6 +19,7 @@ import javax.annotation.Nullable;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 
 public class EntryIndex implements JarIndexer {
 	private final EntryTree<EntryMapping> tree = new HashEntryTree<>();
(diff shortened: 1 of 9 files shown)


enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/jar/EntryIndex.java
```diff
@@ -67,7 +70,29 @@ public boolean hasField(FieldEntry entry) {
 		return this.fieldDefinitions.containsKey(entry);
 	}
 
+	/**
+	 * Checks the entry has been indexed and therefore exists in the JAR file.
```
IotaBread, line 74:
```suggestion
	 * Checks whether the entry has been indexed and therefore exists in the JAR file.
```

enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/jar/EntryIndex.java
```diff
@@ -67,7 +70,29 @@ public boolean hasField(FieldEntry entry) {
...
+	 * @param entry the entry to check
+	 * @return whether the entry exists
+	 * @see #hasEntry(Entry, EnigmaProject)
+	 */
 	public boolean hasEntry(Entry<?> entry) {
+		return this.hasEntry(entry, null);
+	}
+
+	/**
+	 * Checks the entry has been indexed and therefore exists in the JAR file.
```
IotaBread, line 87:
```suggestion
	 * Checks whether the entry has been indexed and therefore exists in the JAR file.
```

enigma/src/main/java/org/quiltmc/enigma/impl/translation/mapping/MappingsChecker.java
```diff
@@ -84,26 +87,42 @@ public Dropped dropEmptyMappings(ProgressListener progress) {
...
+		// account for child mappings that have been dropped already
+		if (!children.isEmpty()) {
+			for (Entry<?> child : children) {
+				if (!dropped.getDroppedMappings().containsKey(child) && !this.hasChildren(child, dropped)) {
+					return true;
+				}
+			}
+		}
+
+		return children.isEmpty();
```
IotaBread, line 123: I feel like the name of this method is the opposite of what this [line] does?

enigma-cli/src/test/resources/drop_invalid_mappings/expected/EmptyMappings.mapping
IotaBread, line 1: Shouldn't this file be deleted?

ix0rai: i prefer having the expected result stored in a file instead of hardcoded. idk

enigma/src/main/java/org/quiltmc/enigma/impl/translation/mapping/MappingsChecker.java
ix0rai: oh you're so right lol

PR #237, opened 2024-11-02
Fix navigator

PiTheGuy:
My last PR broke it


ix0rai (approved):
this is a good change nonetheless, but I'm not entirely sure what the bug being fixed here is. i've never seen the index go below 0 in my mapping

PiTheGuy:
Yeah my previous PR broke it. If you mapped multiple parameters in a row, it would keep decrementing and go below 0

ix0rai:
got it, i see the reproduction

PR #240, opened 2024-11-09
fallback name proposal and name proposal save bypass

ix0rai:
implements #239 

- adds a `bypassValidation()` boolean in `NameProposalService`
  - this boolean allows that service to ignore the regular checks performed on proposed mappings (validating that they either remove a mapping or provide a proposed mapping)
  - this is essential for part of https://github.com/QuiltMC/quilt-mappings/issues/660, since the mappings proposed for classes to apply moj packages have to be `TokenType.DEOBFUSCATED`
- adds a `fallback()` boolean to `NameProposalService`
  - fallback names will not count towards stats by default
  - fallback names will have a different colour in the UI
  - fallback names do not have their own token types, and are expected to be less reliably high-quality names than normal proposed mappings

PR #245, opened 2024-12-08
extra proposal context

ix0rai:
based on #240, required for https://github.com/QuiltMC/quilt-mappings/issues/660

- adds the context of an `Enigma` object for both proposal methods
  - `EntryRemapper` now exposes an `Enigma` to facilitate this
- adds new `readMappings` API in `Enigma`
- replaces `NameProposalService#bypassValidation` with a `validateProposedMapping` method that can be overridden to achieve the same effect
  - this also fixes dynamic proposed names not receiving this validation
    - which will break plugins that abuse that oversight. such as enigma's builtin plugin (i had to fix it). this is now unit tested oops
- improve unit testing to account for bugfixes


IotaBread (approved):
looks great!

PR #248, opened 2025-01-07
fix drop invalid taking 2 runs to resolve a certain drop

ix0rai:
exactly the title. this was noticed on `valueOf` methods on QM (ex. this drop requires two runs: https://github.com/QuiltMC/quilt-mappings/pull/683/files#diff-c9f1603baaaaea5f84c409bf867b7516890ec2f814826189f521eb065887ca67)

this is likely an overengineering, as i thought when starting this we already had sorting implemented. either way, i like the new structure better


enigma-cli/src/test/resources/drop_invalid_mappings/expected/Enums.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS a Gaming
+	FIELD a wawa I
```
IotaBread, line 2: awawa

IotaBread (approved):

PR #250, opened 2025-01-16
Fix deobfuscation icons not updating after initial stat generation completes

PiTheGuy:
Previously they wouldn't update until you clicked on the docker. Now they update immediately


ix0rai (commented):
looks great, i'll test soon!

ix0rai:
@PiTheGuy are you ok with me hijacking your PR and reimplementing this? it has the major issues that this only applies to the all classes docker
and to fix that, i'll have to make it run over all class tree dockers and then add a couple dozen lines of code to make sure that only visible dockers get reloaded

PiTheGuy:
Go ahead. I'm not familiar enough with the codebase to attempt something like that but if you want to, go for it!

ix0rai:
when you're the one who wrote all the docker code in the first place it's a 15 min job lol

PR #254, opened 2025-01-20
Fix methods obfuscated as "new" not creating their tokens

### LINKED ISSUE #249: Issue with obfuscated methods "new"
When using Vineflower or CFR, obfuscated methods with the name `new` are not recognized as obfuscated therefore they can't be renamed. With Procyon or Bytecode they work as expected, but the renames don't carry over when switching back. This doesn't happen for fields.

Images for example's sake.

Image

Image

IotaBread:
Fixes #249. Im not quite sure why, but both cfr and vineflower would ignore methods named "new" while generating tokens for the decompiled source; removing said behavior didn't have any side effects as far as i can tell

tests still needed

Also, debug tokens didn't work, so i fixed those too

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -1,1 +1,4 @@
-
+java {
+	sourceCompatibility = JavaVersion.VERSION_17
+	targetCompatibility = JavaVersion.VERSION_17
+}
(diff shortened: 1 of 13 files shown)


ix0rai (approved):
love getting rid of pointless hardcoding

PR #258, opened 2025-02-15
Fix bug in ClassProvider

### LINKED ISSUE #257: `ClassProvider.getClasses` incorrectly includes subclasses of similarly named classes, causing Vineflower decompilation issues
The class `net.minecraft.client.resources.model.ModelBaker` in the newest snapshot (25w07a) appears very broken when opened and decompiled with Vineflower.

Image

After some investigation, I found the issue. Enigma's `ClassProvider.getClasses` returns subclasses of a given class by checking if its name starts with the root class's name. However, another class named `ModelBakery` is in the same package as `ModelBaker`. This output does get filtered for only classes that contain `$`, so `ModelBakery` itself isn't caught, but its subclasses are. This results in the token highlighter having tokens from the subclasses of `ModelBakery` mixed in with tokens from `ModelBaker`. The token highlighter is attempting to highlight garbage tokens taken from an unrelated class, which causes it be all broken.

PiTheGuy:
Fixes #257

### DIFF
--- a/enigma/src/main/java/org/quiltmc/enigma/api/class_provider/ClassProvider.java
+++ b/enigma/src/main/java/org/quiltmc/enigma/api/class_provider/ClassProvider.java
@@ -32,8 +32,12 @@ public interface ClassProvider {
 		String finalClassName = className;
 		return this.getClassNames()
 				.stream()
-				.filter(s -> s.lastIndexOf('/') == depth && s.startsWith(finalClassName))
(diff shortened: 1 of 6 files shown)


ix0rai (approved):
thank you so much for tracking down this bug! i would appreciate if you could make a unit test for it :)

ix0rai (approved):
unit test looks great!

PR #259, opened 2025-02-15
Minor wording changes to the discard mappings dialog

PiTheGuy:
Addresses some inconsistencies in the discard mappings dialog


ix0rai (approved):
thanks!

PR #266, opened 2025-02-28
add a search bar to the structure panel

### LINKED ISSUE #263: [Feature Request] allow searching for class/field/method declarations within the current file
The current search dialog allows searching through the entire jar, but sometimes I'm looking for a specific method in the current file. The search dialog or even the ctrl+f search are not convenient when looking for a specific field/method (or even inner class sometimes) in the current file if it's very long.

- ctrl+f just matches the string, which includes calls to the desired field/method/class and even entirely different entries if the search key is part of their name. 
- ctrl+shift+f searches for declarations through the whole jar, and if you're searching for a method that is implemented by many classes, or if the field/method name appears in several classes, it's a pain to sift through the results.

A variation of the search dialog that is limited to just the current file would be a convenient solution.

ix0rai:
fixes #263 !

super simple contains search that rebuilds the tree with each key pressed
image


ix0rai:
this is done! it was mentioned on discord that maybe we could allow tree navigation via arrow keys while searching, but i won't be implementing that personally as i couldn't find a clean way to do it.

supersaiyansubtlety (approved):

PR #268, opened 2025-03-08
Fix navigator keybind focus issues

PiTheGuy:
Using Ctrl + Up Arrow to navigate up on the navigator would previously shift focus away from the editor panel and prevent you from mapping or navigating until you manually refocused it. This PR fixes this issue.


ix0rai (approved):
thanks!!

Issue #267, opened 2025-03-08
[Enhancement] Rework the navigator


PiTheGuy:
The current implementation of the navigator has a fundamental flaw. It functions correctly only if navigation is done exclusively through the navigator itself, without manually clicking on mappings.

Currently, when a mapping is completed, the navigator decrements its index to prevent skipping mappings as the total count decreases. Since navigating forward increases the index, this system maintains a consistent position while progressing through mappings.

However, issues arise when using the mouse to manually select mappings. In this case, the navigator’s index does not update accordingly. As a result, when a mapping is completed, the index still decrements, causing the navigator to backtrack. Over time, this behavior leads it to regress all the way to the beginning of the file.

While this issue might be fixable within the current system, I believe a more intuitive approach would be for the navigator to always navigate to the mapping closest to the cursor, rather than maintaining a separate stored position.

PR #269, opened 2025-03-10
Navigator rework

PiTheGuy:
Implements the navigator rework as described in #267. 

- Navigation is now always relative to the cursor rather than maintaining a separate stored position.
- The navigator still wraps around when necessary.
- The navigator's index is still stored but is now only used for display purposes. It is updated accordingly when navigating.


ix0rai:
thanks! i'll look into this soon

PR #270, opened 2025-03-11
fix navigating on click when clicking to exit quick find

supersaiyansubtlety:
Fixes a bug where if you press `ctrl+f` to quick-find, `shouldNavigateOnClick` remains `true` from when you held `ctrl`.

This means that you navigate to any token you click on after opening quick-find, even if not holding control.


ix0rai (approved):
quite honestly never realised this was a bug. thanks for the fix!

PR #276, opened 2025-05-21
Autosave mappings

PiTheGuy:
- Automatically save mappings when a change is made
- Saving is done in the background and is shown in the progress bar
- Can be disabled in the config


ix0rai (approved):
looks great, thanks!

PR #278, opened 2025-08-13
Drop proposed mappings from the files

OroArmor:


enigma/src/main/java/org/quiltmc/enigma/api/EnigmaProject.java
```diff
@@ -167,6 +169,68 @@ private Collection<Entry<?>> dropMappings(EntryTree<EntryMapping> mappings, Prog
...
+//			ParentedEntry<?> parent = ((ParentedEntry<?>) proposedMapping.getEntry());
+//			while (parent != null && parent.getParent() != null) {
+//				parent = (ParentedEntry<?>) parent.getParent();
+//			}
+//
+//			if (!parent.getFullName().equals("net/minecraft/unmapped/C_ygpuayyc")) {
+//				continue;
+//			}
+
+			if (proposedMapping.getValue().sourcePluginId().equals("quiltmc:name_proposal/delegate_params")) {
```
supersaiyansubtlety, line 214: this should preserve fallback-proposed names as well
also why do delegate params need an exception?

OroArmor: I believe this was a work around until we figure out why we are getting the massive spooky diffs,

enigma/src/main/java/org/quiltmc/enigma/api/translation/mapping/tree/MergedEntryMappingTree.java
```diff
@@ -42,10 +43,15 @@ public EntryMapping remove(Entry<?> entry) {
 	@Override
 	public EntryMapping get(Entry<?> entry) {
 		EntryMapping main = this.mainTree.get(entry);
+
 		if (main == null || (main.equals(EntryMapping.OBFUSCATED) && this.secondaryTree.contains(entry))) {
 			return this.secondaryTree.get(entry);
 		}
 
+		if (main.tokenType().equals(TokenType.OBFUSCATED) && this.secondaryTree.contains(entry)) {
```
OroArmor, line 51: @ix0rai what is your opinion on this? it allows for javadocs to be on empty mappings, but get a name from a proposer

enigma/src/main/java/org/quiltmc/enigma/api/EnigmaProject.java
supersaiyansubtlety: ah, ok
it could be a useful to have an arg for proposer ids to exclude

OroArmor: potentially, but i also feel like that defeats a bit of the purpose of this feature, which is to make sure that no more proposed names are left in the mappings. If the proposer is fragile enough to break with this, its probably not written very well.

supersaiyansubtlety: true
I wouldn't expect it to be used often
maybe it would be better not to have an arg; if a proposer is having issues it can just be set to fallback as a workaround

enigma/src/main/java/org/quiltmc/enigma/api/translation/mapping/tree/MergedEntryMappingTree.java
ix0rai: just making sure you know i'm not ignoring you, i'm pretty busy rn and i want to look into javadoc support on obf mappings a bit deeper

PR #287, opened 2025-08-30
Fix `TestPackageRename`

supersaiyansubtlety:
Call `setupTheme` in `TestPackageRename`; fixes NPE


ix0rai (approved):
perfect! thanks for fixing this.

PR #298, opened 2025-09-14
Fix 201

### LINKED ISSUE #201: Reports wrong type for (some?) lambda parameters
It's reporting type `ItemStack` for `slot` which is an `EquipmentSlot`. 

If you move the cursor to the `stack` param, the type is missing entirely.

It seems like the index is just off by one?

java_t3ld82Xnkc

supersaiyansubtlety:
Fixes #201

`IdentifierPanel::refreshReference` passed deobf `LocalVariableEntry`s to `EntryIndex` which only contains obf entries.

This fixes that and:
- gets the param descriptor from `EntryIndex` instead of deriving it from the parent method
- eliminates the possibility of incorrect types in the `IdentifierPanel` due to index offsets
- shows `<unknown>` for the types of un-indexed locals (this should only happen for non-args)

~~Based on #282 because it adds param indexing.~~ rebased

~~Ready once #282 is merged.~~ ready

### DIFF
--- a/enigma-swing/src/main/java/org/quiltmc/enigma/gui/panel/IdentifierPanel.java
+++ b/enigma-swing/src/main/java/org/quiltmc/enigma/gui/panel/IdentifierPanel.java
@@ -1,10 +1,9 @@
 package org.quiltmc.enigma.gui.panel;
 
+import org.quiltmc.enigma.api.EnigmaProject;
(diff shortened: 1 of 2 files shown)


ix0rai (approved):
lovely! kinda embarrassing how i never noticed that this was the fix. thanks for squashing this bug!

PR #300, opened 2025-09-17
Auto save improvements

PiTheGuy:
- Auto save code was moved to a lower level, meaning it now applies more broadly
- Auto save is now deferred for package renames


supersaiyansubtlety (approved):
looks good

ix0rai (approved):
looks good, thanks!

PR #303, opened 2025-09-22
Improve keybinds

supersaiyansubtlety:
Uses `InputMap`s+`ActionMap`s instead of `KeyListener`s for `KeyBind`s.

Also:
- fixes a bug where changing a keybind via the ui would add a new keybind instead of replacing the old one (`isNewCombination` was inverted)
- simplifies `EditorPanel` ctrl+click to navigate logic
- changes default `EDITOR_SHOW_CALLS` binding to `ctrl + alt + c` because `ctrl + c` is copy on windows (before it was both copying and showing calls)


ix0rai (approved):
looks great! this is giving me a few ideas for other improvements to keybinds:
- improve the logic of the keybinding menus
  - when editing an individual keybind, "save" does not save to config. the keybinding will only be saved when the *other* save button (same text!) is pressed in the top-level menu. very unintuitive
  - "save" is on the left, with no visual differentiation. it's common in UI to put the continue button on the left, but usually with accent colour in order to show which to press. we should look at adding colour like what is seen on the "save as" dialog: Image
- minor bug: you can create multiple of the exact same keybind (if the keybind is ctrl+shift+f, i can just add a second ctrl+shift+f without issue)
- the config file is unreadable:

```toml
[key_codes]
	search = ["70;c0", "70;c0"]
	search_field = ["76;c0"]
```
we should look at either adding comments to describe what the bindings actually are or changing the encoding to something human-readable. 

ANYWAY. i'm just yapping i'll break all that out into separate issues

Issue #274, opened 2025-04-18
Interface implementers with external method implementations prevent naming in GUI


supersaiyansubtlety:
To reproduce:
- checkout https://github.com/QuiltMC/quilt-mappings/commit/7847ea68c2de072b9f283efda3e36b861f729ba3
- run `gradlew mappingsUnpicked`
- open `AbstractNbtList` in the GUI
- show implementations of `m_fldtrget`, note that `NbtList`'s implementation is named because this method is its implementation of `java.util.AbstractList::remove`
- try to name `m_fldtrget` -> `remove`
- see: ![Image](https://github.com/user-attachments/assets/57384a25-1053-4f65-b6a8-443bee9a8b38)

I can give it the correct name by editing the `.mapping` file, and the GUI show it if I reload mappings afterward.

Ideally Enigma would recognize cases like this and both give the correct name to method in the parent interface (`AbstractNbtList` in this case) and prevent renaming the method.

supersaiyansubtlety:
The renamable part is fixed on <https://github.com/supersaiyansubtlety/enigma/tree/fix-274>, but it requires indexing `java.*` classes which results in ASM throwing `AnalyzerException`s when I try using it on QMap.

PR #306, opened 2025-09-26
Fix 91

### LINKED ISSUE #91: [bug] enigma does not consider that classes all override `java/lang/Object`
Since enigma is not aware that all classes extend `Object`, bugs like https://github.com/fabricmc/enigma/issues/140 are caused.
[cited issue being that a method with no parameters that returns a string can be named toString, causing conflicts when the mappings are used]

supersaiyansubtlety:
fixes #91

repurposed from my fix for #274 that was plagued by `AnalyzerException`s when used by QMap

this:
- exposes `Enigma`'s lib index
- adds and exposes a combined index containing both lib and main jar entries to `Enigma`
- passes the combined index to `EntryRemapper` instead of the main jar index
- disallows renaming entries that trace back to libs (effects #274, but only for the limited number of lib classes currently indexed)
- removes `java/lang/Object`'s explicit exclusion from `InheritanceIndex` (added in [this commit](<https://github.com/QuiltMC/enigma/commit/c0b0388aae76d39d780c2b4ad2531feb9d633e62#diff-5e3f094afadeceb89ba971157f77139e7c7b7837c3b7c89b26cac6ca2e4299a1R41>))

Also adds but does not use `JavaClassProvider` which can be swapped for `ClasspathClassProvider` to fix more of #274 once `AnalyzerException`s are sorted.

When squash-merging, be sure to change the commit message about fixing #274 so it doesn't get prematurely closed.

### DIFF
--- a/enigma/src/main/java/org/quiltmc/enigma/api/EnigmaProject.java
+++ b/enigma/src/main/java/org/quiltmc/enigma/api/EnigmaProject.java
@@ -121,14 +121,23 @@ public class EnigmaProject {
 		return this.classProvider;
 	}
 
+	/**
+	 * Gets the index of the main jar of this project; the jar being mapped.
+	 */
 	public JarIndex getJarIndex() {
 		return this.jarIndex;
 	}
 
+	/**
+	 * Gets the index of the library jars of this project.
+	 */
 	public JarIndex getLibIndex() {
 		return this.libIndex;
 	}
 
+	/**
+	 * Gets the index of the main jar <em>and</em> library jars of this project.
+	 */
 	public JarIndex getCombinedIndex() {
 		return this.combinedIndex;
 	}
--- a/enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/jar/CombinedJarIndex.java
+++ b/enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/jar/CombinedJarIndex.java
@@ -1,10 +1,14 @@
 package org.quiltmc.enigma.api.analysis.index.jar;
 
+import org.quiltmc.enigma.api.EnigmaProject;
 import org.quiltmc.enigma.api.class_provider.ProjectClassProvider;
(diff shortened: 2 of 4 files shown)


enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/jar/CombinedJarIndex.java
```diff
@@ -0,0 +1,42 @@
+package org.quiltmc.enigma.api.analysis.index.jar;
+
+import org.quiltmc.enigma.api.class_provider.ProjectClassProvider;
+import org.quiltmc.enigma.impl.analysis.index.AbstractJarIndex;
+
+import java.util.Collection;
+
+public class CombinedJarIndex extends AbstractJarIndex {
```
ix0rai: would appreciate a line or two of top-level docs here!

enigma/src/main/java/org/quiltmc/enigma/api/EnigmaProject.java
```diff
@@ -56,22 +60,25 @@ public class EnigmaProject {
...
 	private final JarIndex libIndex;
+	private final JarIndex combinedIndex;
 	private final byte[] jarChecksum;
+	private final Map<MethodEntry, Boolean> libraryMethodOverrideCache = new HashMap<>();
 
 	private EntryRemapper remapper;
 	private MappingsIndex mappingsIndex;
 
-	public EnigmaProject(Enigma enigma, Path jarPath, ClassProvider classProvider, JarIndex jarIndex, JarIndex libIndex, MappingsIndex mappingsIndex, EntryTree<EntryMapping> proposedNames, byte[] jarChecksum) {
+	public EnigmaProject(Enigma enigma, Path jarPath, ClassProvider classProvider, JarIndex jarIndex, JarIndex libIndex, JarIndex combinedIndex, MappingsIndex mappingsIndex, EntryTree<EntryMapping> proposedNames, byte[] jarChecksum) {
```
ix0rai: parameter situation here is becoming critical

Issue #140, opened 2023-08-16
Show javadocs/other information on hover


OroArmor:
Showing javadocs on hover would be really helpful for reading what functions do. Even if there are no docs, it would be helpful to see what the parameter names are (and somehow if they are mapped) and what the return type is.

It would also allow seeing library names for parameters and understand what the code for them is doing, especially if libraries had their javadocs indexed.

PR #307, opened 2025-09-28
Entry tooltips

supersaiyansubtlety:
Adds  #140

~~based on #304~~ rebased

---

### Notes:
- There's one weird issue I don't know what to make of: selecting javadoc text breaks the text rendering. They're just `JTextArea`s. I've tried using `JEditorPanel`s instead, and I've tried disabling the drag-to-move feature; neither helped.
- There's one TODO that will require a followup PR (#317): declaration snippets should include the original source's line numbers; requires line number offset implementation in syntaxpain.
- This does not use `JToolTip` or `ToolTipManager` because I couldn't get a custom `JToolTip` to properly move+resize when the mouse moved to a new token. Instead it uses a custom `JWindow` for the tooltip and several `Timer`s to manage it.
- The tooltip size is limited to half the screen width/height with an additional 600px (scaled) hard cap on width. All tooltip content except the parent label is in a scroll pane which fills available space. If there's not enough room for everything, the scroll pane gets automatically scrolled to the bottom so the declaration snippet is in view. Unfortunately this means the top portion of javadocs are scrolled off the top. Ideally the javadocs and the snippet would have separate scroll panes (the snippet already has one from `BaseEditorPanel`), but I couldn't get two scroll panes to resize smartly in limited space (at least in a `GridBagLayout`, used to expand to fill width).

---

### Improvements to `RecordGetterFindingVisitor`:
- renames it `RecordIndexingVisitor`
- adds map of record classes to component fields
- adds map of record classes to component getters
- makes its results accessible
-  fixes a bug preventing it from finding inner records' getters because it used `getName()` instead of `getFullName()`

---

### Also:
- makes navigating to a record component getter declaration with no explicit declaration navigate to its field instead
- eliminates `FeaturesSection` by moving `enableClassTreeStatIcons` to `StatsSection` and `autoSaveMappings` to `EditorConfig`

---
### The tooltip:

The tooltip shows the parent, javadocs, and declaration snippet of the entry under the cursor.

enigma-entry-tooltip

Clicking the parent label or tokens within the declaration snippet updates the tooltip with that entry's info; control-clicking navigates to the entry. Control-clicking a top-level class' package navigates to that package. The tooltip may be moved by dragging from anywhere that doesn't have other click functionality.

---

<details>

<summary><h3>Config</h3></summary>

Tooltips and their interact-ability can each be disabled; configs are in a new `editor.toml` file. This moves #304's persistent config to `editor.toml`.

`editor.toml`:
```toml
# Enables auto save functionality, which will automatically save mappings when a change is made.
# default: false
auto_save_mappings = false
# Whether editors' quick find toolbars should remain visible when they lose focus.
# default: true
persistent_quick_find = true

# Settings for editors' entry tooltips.
[entry_tooltips]
	# Whether tooltips are enabled.
	# default: true
	enable = true
	# Whether tooltips can be clicked and interacted with to navigate their content.
	# default: true
	interactable = true
```

enigma-entry-tooltips-config-gui

</details>

---

### Tasks:
- [x] obtain entry under cursor
- [x] account for entry type
- [x] account for entry source (current class/other class)
- [x] show declaration of entry in source
- [x] de-indent trimmed declaration source
- [x] add javadoc if present
- [x] add entry parent name
- [x] make tooltips disablable
- [x] make tooltips optionally clickable
- [x] allow clicking entries inside tooltip to update the tooltip with that entry
- [x] polish
- [x] merge upstream once #304 is merged (both mess with `EditorPanel`)
- [x] merge upstream once #308 is merged (both add javaparser and `LineIndexer`)

~~add "bread crumbs" back button to go back to previous tooltip contents~~ not tooltip-y

### DIFF
--- a/enigma-swing/src/main/java/org/quiltmc/enigma/gui/panel/DeclarationSnippetPanel.java
+++ b/enigma-swing/src/main/java/org/quiltmc/enigma/gui/panel/DeclarationSnippetPanel.java
@@ -40,6 +40,7 @@ import org.quiltmc.enigma.api.translation.representation.entry.MethodEntry;
 import org.quiltmc.enigma.gui.Gui;
 import org.quiltmc.enigma.gui.config.Config;
 import org.quiltmc.enigma.gui.highlight.BoxHighlightPainter;
+import org.quiltmc.enigma.util.I18n;
 import org.quiltmc.enigma.util.LineIndexer;
 import org.quiltmc.enigma.util.Result;
 import org.quiltmc.syntaxpain.LineNumbersRuler;
@@ -77,7 +78,7 @@ public class DeclarationSnippetPanel extends BaseEditorPanel {
 			if (!this.isBounded()) {
 				// the source isn't very useful if it couldn't be trimmed
 				// set this text so it doesn't waste space or cause confusion
-				this.editor.setText("// Unable to locate declaration");
+				this.editor.setText("// " + I18n.translate("editor.snippet.message.no_declaration_found"));
 				this.editor.getHighlighter().removeAllHighlights();
 			} else {
 				this.resolveTarget(source, target)
--- a/enigma/src/main/resources/lang/en_us.json
+++ b/enigma/src/main/resources/lang/en_us.json
@@ -157,6 +157,7 @@
 	"editor.tooltip.label.from": "from",
 	"editor.tooltip.label.inherited_from": "inherited from",
 	"editor.tooltip.message.no_source": "No source available",
+	"editor.snippet.message.no_declaration_found": "Unable to locate declaration",
 
 	"info_panel.identifier": "Identifier Info",
 	"info_panel.identifier.none": "No identifier selected",


enigma/src/main/java/org/quiltmc/enigma/impl/plugin/RecordIndexingVisitor.java
```diff
@@ -72,15 +88,22 @@ public MethodVisitor visitMethod(final int access, final String name, final Stri
...
 		try {
-			if (this.clazz != null) {
-				this.collectResults();
-			}
+			this.collectResults();
+
+			this.clazz = null;
+			this.recordComponents.clear();
+			this.fields.clear();
+			this.methods.clear();
```
supersaiyansubtlety: please double-check that moving the clearing here and clearing fields+methods won't break anything

supersaiyansubtlety:
ready!

enigma/src/test/java/org/quiltmc/enigma/input/z_tooltip/Constructors.java
```diff
@@ -0,0 +1,21 @@
...
+public class Constructors {
+	public Constructors(String outerArg) {
+		new Constructors("outer arg");
+
+		new Methods() {
+			// multiline declaration to test unindenting
+			@Deprecated
+			@Override
+			void abstraction() {
+				// tests #252
```
ix0rai: not a comment on this PR but i'm going to make sure links like this have archived versions attached, i don't want any information lost if we ever have to move off github

enigma-swing/src/main/java/org/quiltmc/enigma/gui/panel/EntryTooltip.java
```diff
@@ -0,0 +1,845 @@
...
+
+		final Font editorFont = ScaleUtil.scaleFont(Config.currentFonts().editor.value());
+		final Font italEditorFont = ScaleUtil.scaleFont(Config.currentFonts().editor.value().deriveFont(Font.ITALIC));
+
+		int gridY = 0;
+
+		{
+			final Box parentLabelRow = Box.createHorizontalBox();
+
+			final JLabel from = labelOf(inherited ? "inherited from" : "from", italEditorFont);
```
ix0rai: would be nice if this was translatable! doesn't need retranslation logic, but make sure it uses formatting (%s) to account for different semantics

ix0rai: i agree, it should!

PR #308, opened 2025-09-28
Fix navigation to method calls within static initializers

PiTheGuy:


enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/EnigmaTextTokenCollector.java
```diff
@@ -80,73 +108,273 @@ public void addTokensToIndex(SourceIndex index, UnaryOperator<Token> tokenProces
...
+		TextRange textRange = new TextRange(this.positionToIndex(range.begin), this.positionToIndex(range.end));
+		this.classRanges.put(getClassEntry(name), textRange);
+		decl.getMembers().forEach(member -> {
+			if (member instanceof TypeDeclaration<?> child) {
+				this.addClassAndChildren(child, name + "$" + child.getNameAsString());
+			}
+		});
+	}
+
+	private int positionToIndex(Position position) {
```
supersaiyansubtlety, line 229: I think it'd be better to greedily build a `List<Integer> indexesByLine` in `parseSource`.

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/EnigmaTextTokenCollector.java
```diff
@@ -80,73 +108,273 @@ public void addTokensToIndex(SourceIndex index, UnaryOperator<Token> tokenProces
...
+		ParseResult<CompilationUnit> parseResult = new JavaParser(config).parse(this.content);
+		if (!parseResult.isSuccessful()) {
+			Logger.warn("Failed to parse source: {}", parseResult.getProblems());
+			return;
+		}
+
+		CompilationUnit unit = parseResult.getResult().get();
+		List<InitializerDeclaration> initializers = unit.findAll(InitializerDeclaration.class, InitializerDeclaration::isStatic);
+		for (InitializerDeclaration decl : initializers) {
+			Range range = decl.getRange().orElseThrow(() -> new IllegalStateException("No range for initializer"));
```
supersaiyansubtlety, line 125: In a brief look into javaparser I wasn't able to find a guarantee that ranges will be present.
If that's the case, I think an error should be loggerd and the declaration skipped if it has no range.

supersaiyansubtlety (changes requested):
Doesn't work as expected here:
java_3gNZFWzbX2

- the `b.b()V` call doesn't navigate; it should got to the runnable lambda
- the `b.<clinit>()V` call navigates to the runnable lambda; it should go to the static init call
- the `b.<init>()V` call works correctly

supersaiyansubtlety (approved):

ix0rai (changes requested):
i may be doing something wrong but i was unable to get this to work for lambdas. using enigma's built in `:lambdaTestGui`, navigating to the top call (in the top level of the method) works, but as before, the two calls inside the lambda and nested lambda do not navigate properly.
image

however, this does fix the issue for static init!

ix0rai (approved):
looks great! this is a huge fix, i'm so happy to get it in.
i'd appreciate if you could add some unit testing for this, but if not i can always just open an issue and get to it when i have time. thanks for the PR!

PiTheGuy:
I think I'll let you handle the unit tests for this one.

ix0rai:
sounds good!

Issue #323, opened 2025-10-28
Quick paste functionality no longer works


PiTheGuy:
A recent change (likely #303) broke the ability to use Ctrl + V to quickly paste the contents of your clipboard as the name of the selected entry. The Paste Text menu item still works but the keyboard shortcut for it no longer functions properly.

supersaiyansubtlety:
fixed by #324

PR #324, opened 2025-10-29
Fix 323

supersaiyansubtlety:
Removes a `JEditorPane` UI input map conflict with `EDITOR_PASTE`, fixing #323

Also makes conflict removal more aggressive by checking all focus conditions and input map parents.


enigma-swing/src/main/java/org/quiltmc/enigma/gui/util/GuiUtil.java
```diff
@@ -344,10 +345,16 @@ public enum FocusCondition {
 		 */
 		WHEN_ANCESTOR_OF_FOCUSED_COMPONENT(JComponent.WHEN_ANCESTOR_OF_FOCUSED_COMPONENT);
 
+		public static final ImmutableList<FocusCondition> VALUES = ImmutableList.copyOf(values());
```
ix0rai: why is this needed?

ix0rai (approved):
looks great, thanks for the quick fix!

PR #326, opened 2025-11-03
Improve number input

### LINKED ISSUE #321: Improve number input dialogs
2.7's [`ScaleMenu`](https://github.com/QuiltMC/enigma/blob/develop/2.7/enigma-swing/src/main/java/org/quiltmc/enigma/gui/element/menu_bar/view/ScaleMenu.java#L70) and #318's [`IntRangeConfigMenuItem`](https://github.com/supersaiyansubtlety/enigma/blob/improve-selection-highlighting/enigma-swing/src/main/java/org/quiltmc/enigma/gui/element/IntRangeConfigMenuItem.java#L80) both use `JOptionPane.showInputDialog(...)` to create a dialog prompting the user for a number.

The dialog accepts any string and gives no feedback when the input is invalid.

It would be much  more user-friendly to create custom dialogs that don't allow submitting invalid input and give feedback on what's invalid.

Polish could include:
- outlining the input field in red when it's invalid
- displaying a red message explaining why the input is invalid
  - not a [whole] number
  - not in range
- step buttons

supersaiyansubtlety:
adds #321

~~based on #318~~ rebased

Replaces the `JOptionPane` dialogs opened by `View > Scale > Custom (N%)...` and `View > Selection Highlight > Blink delay (Nms)...` with custom `NumberInputDialog`s.

---

<h4><code>NumberInputDialog</code>s:</h4>

- only allow submitting ints/floats within a specified range
- display a red border and error message when the current input is invalid
- play the OS's error chime when invalid input is entered
- have step up/down buttons
- have key binds for cancel, submit, step up/down, and alt step up/down

---

<h4>Also:</h4>

- gives the custom scale dialog a more informative message
- displays any current custom scale in the `Custom...` menu item
- adds `@RangedFloat` to `Config.main().scaleFactor` and reduces the max to 5 (from 10)
- eliminates `ScaleChangeListener` and `ScaleUtil::setScaleFactor`; `TrackedValue` methods make them unnecessary

---

<h4>Notes:</h4>

- uses custom `NumberTextField`s instead of `JFormattedTextField`s because the `Format`s used by `JFormattedTextField` are too lenient: they seem to ignore all trailing characters after valid input (including `.` after ints). The `NumberTextField`s use `Integer::parseInt`/`Float::parseFloat`; the only added leniency is trimming leading/trailing white space.
- I considered trying to disallow entering characters that couldn't lead to valid input, but I think it could be counter intuitive (e.g. a user tries to paste a negative number at the end of the current input; it fails because a `-` cannot be mid-number), and it would require a custom `Document` which would probably require magical toolkit installation and generally increase complexity.
- I added `menu.view.scale.custom.selected` translations for French and Japanese based on `menu.view.scale.custom` assuming the formatting would be the same as in English; I don't speak French or Japanese

---

<h4>Custom Scale:</h4>

| initial | max |
|-|-|
| scale-dialog-initial-updated | scale-dialog-max-default-button |

| low | NaN | 
|-|-|
| scale-dialog-low-updated | scale-dialog-nan-updated |

---

<h4>Blink Delay:</h4>

| initial | min |
|-|-|
| blink-dialog-initial-updated | blink-dialog-min-default-button |

| high | NaN |
|-|-|
| blink-dialog-high-updated | blink-dialog-nan-updated |


supersaiyansubtlety:
ready!

enigma/src/main/resources/lang/en_us.json
```diff
@@ -78,11 +78,13 @@
 	"menu.view.selection_highlight.blink_delay": "Blink delay (%sms)...",
 	"menu.view.selection_highlight.blink_delay.dialog_title": "Blink Delay",
 	"menu.view.selection_highlight.blink_delay.dialog_explanation":
-	"The milliseconds to blink on and off highlighting that indicates an entry that has been navigated to.",
+	"The time (in milliseconds) highlighting indicating which entry has been navigated to blinks on/off.",
```
eristhea:
```suggestion
	"The time (in milliseconds) that an entry's highlight will blink on and off to indicate that it's been navigated to.",
```
i think this is easier to understand, a bit better for the logical flow of information

eristhea (approved):
looks good! i just want to bikeshed the new wording a little

Issue #325, opened 2025-11-01
error handling for plugins


ix0rai:
plugins should be able to declare which versions of enigma they're compatible with to prevent untraceable errors like #265

PR #328, opened 2025-11-05
Plugin error handling

supersaiyansubtlety:
adds #325

Uses compile-time int constants for a default `EnigmaPlugin::supportsEnigmaVersion` that matches versions with the same major and minor version as the version compiled against.

Still needs a test making sure that actually works (copy `:engima`, bump copy's version, implement `EnigmaPlugin` and ensure it throws when run with copy).

### DIFF
--- a/enigma/src/main/java/org/quiltmc/enigma/util/Version.java
+++ b/enigma/src/main/java/org/quiltmc/enigma/util/Version.java
@@ -40,7 +40,7 @@ public record Version(int major, int minor, int patch) implements Comparable<Ver
 	 * @param minor the minor number part
 	 * @param patch the patch number part
 	 *
-	 * @throws IllegalArgumentException if {@code major}, {@code minor}, or {@code patch} is negative
+	 * @throws IllegalArgumentException if the passed {@code major}, {@code minor}, or {@code patch} is negative
 	 */
 	public Version {
 		validateNonNegative(major, "major");


enigma/src/main/java/org/quiltmc/enigma/api/Enigma.java
```diff
@@ -62,6 +63,13 @@ public class Enigma {
 	public static final String URL = "https://quiltmc.org";
 	public static final int ASM_VERSION = Opcodes.ASM9;
 
+	// These are compile-time constants.
```
ix0rai: why are these constant instead of being based on the existing `version.properties`? i can see that it's a deliberate choice, but i'm not clear on the logic behind it

supersaiyansubtlety: I might be derping, but iiuc:
- compile time constants will get inlined, so plugins using the default `supportsEngimaVersion` will "remember" the version they compiled against
- the `VERSION` string is not a compile time constant, so if the default `supportsEngimaVersion` used it, it would always match the run time version, making it pointless

supersaiyansubtlety: Actually I think I'll update that to use `Integer::toString` on the ints instead of `Integer::parseInt` on the strings; a little stricter.

supersaiyansubtlety:
This doesn't work like I thought.
If the plugin doesn't override `supportsEnigmaVersion`, it uses the implementation from the currently running enigma, making the inlined constants pointless.

PR #329, opened 2025-11-07
Single class reloads

PiTheGuy:
Adds a button to the class selector context menu to reload mappings for the selected class


enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/mapping/MappingsIndex.java
```diff
@@ -72,6 +72,23 @@ public void indexMappings(EntryTree<EntryMapping> mappings, ProgressListener pro
...
+		this.indexNodes(nodes, true);
+	}
+
+	public void reindexClass(ClassEntry classEntry, EntryTree<EntryMapping> oldMappings, EntryTree<EntryMapping> newMappings, ProgressListener progress) {
+		this.progress = progress;
+		oldMappings.remove(classEntry);
+
+		EntryTreeNode<EntryMapping> classNode = newMappings.findNode(classEntry);
+		if (classNode == null) {
+			throw new IllegalArgumentException("Class entry not found in mappings");
```
supersaiyansubtlety: could you include the missing class entry's name in the message?

supersaiyansubtlety (changes requested):
I got this crash when I tried to use the menu item: [enigma_crash-329.log](https://github.com/user-attachments/files/23427621/enigma_crash-329.log)

PR #330, opened 2025-11-08
Plugin error handling (but this time it works)

supersaiyansubtlety:
Like #328 but with no default implementation of `supportsEnigmaVersion`.

I kept the compile-time major/minor/patch constants so plugins can use them in their implementations.

I also finished the test for the 'recommended' implementation that uses the constants and added a test that 'simulates' running a pre-2.7 plugin on 2.7+.

The 'simulation' just uses a copy of 2.6.2's `EnigmaPlugin` and an empty `EnigmaPluginContext`.

The tests are both in the build script.


ix0rai (approved):
looks amazing! the amount of work put into the test is huge, and i'm sure it'll be useful for future testing as well.

PR #331, opened 2025-11-08
Rewrite result

supersaiyansubtlety:
Rewrites `Result` so it has separate `Ok` and `Err` implementations.

Also adds javadocs.

~~Notably it doesn't hide the `Ok` and `Err` implementations; it didn't seem necessary.~~ hid them in package-private classes to be safe
(I'm also not sure if `util` is part of the api or not)


ix0rai:
to me `util` isn't API, we're allowed to break whenever we want. if a util class is worth exposing as API we'll do an `api.util` package

PR #332, opened 2025-11-09
Fix window pos

### LINKED ISSUE #272: broken configs can cause UI to silently fail
Following the `java -jar` instructions in `README.md` does not work as expected for any release in the past year or so. None of the artifacts past version 2.0 launch the user interface.

Using `java -jar enigma-swing-all.jar`:
- `1.X`: Works as expected
- `2.0`: Crashes with IOOBE in `org.quiltmc.enigma.gui.config.Config.getMostRecentProject(Config.java:131)`
- `2.1+`: Creates an empty window that can't be interacted with _(Can't maximize, bring to front, has no title, has no preview when hovering over the windows dock bar icon)_

I've tried launching with Java 17 and 21 but there is no change in behavior. Is there some additional requirement not outlined in the `README.md` ?

supersaiyansubtlety:
~~based on #307 because it adds `Utils::clamp`~~ rebased

- fixes crash when `Vec2i` configs had missing components
- enures the main window is on-screen when opening (probably fixes #272)

---

I was testing #272 and I *think* the `2.1+` behavior described was because the window was off-screen:
> 2.1+: Creates an empty window that can't be interacted with (Can't maximize, bring to front, has no title, has no preview when hovering over the windows dock bar icon)

Though the window preview was blank, not missing or transparent:
enigma-272-config-dock-preview

The config attached to #272 has this:
```toml
# How big the Enigma window will open, in pixels.
# default: Vec2i[x=1024, y=576]
[window_size]
	x = 1936
	y = 1056

# The position the top-left corner of Enigma's window will be the next time it opens, in pixels.
# default: Vec2i[x=0, y=0]
[window_pos]
	x = -32000
	y = -32000
```

### DIFF
--- a/enigma-swing/src/main/java/org/quiltmc/enigma/gui/Gui.java
+++ b/enigma-swing/src/main/java/org/quiltmc/enigma/gui/Gui.java
@@ -238,7 +238,7 @@ public class Gui {
 		final Dimension screenSize = Toolkit.getDefaultToolkit().getScreenSize();
 
 		// keep the whole width (or as much as will fit) on-screen so window controls in the title bar are accessible
-		final int clampedX = Utils.clamp(windowPos.x, 0, Math.max(0, screenSize.width - frame.getSize().width));
+		final int clampedX = Utils.clamp(windowPos.x, 0, Math.max(0, screenSize.width - frame.getWidth()));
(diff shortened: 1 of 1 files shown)


supersaiyansubtlety:
ready!

PR #333, opened 2025-11-09
Internal proposers

### LINKED ISSUE #252: Parameter link lost issue
I'm not sure how to title this issue, so please feel free to add a better name. I don't know the technical name of this kind of nesting.

As you can see from this screenshot of CFR decompiled code (which is pretty much identical to Vineflower in this case, and very similar to Procyon and they all present the same issue), the `string` parameter inside `ResourceFinder` is not recognized where it is then used at lines 144, 152 and 157. The only difference I can see with the similar `resourceName` parameter inside `ResourceCreator` is the fact that the `string` one is also nested inside `Resource` when it is accessed.

Image

### LINKED ISSUE #314: default record component getters should automatically get their component's name
Since default record component getters aren't shown anymore, it can be annoying to name them because they aren't always referenced in the record that declares them.

Furthermore, since their name must always match their component field, they should automatically receive that name and not be manually renamable.
(I believe this will require a kind of internal-only dynamic proposer that ignores normal namability checks.)

The GUI should prevent naming them, and any names for them in mapping files should be dropped as invalid.

Alternatively, the GUI could "allow" naming them, but it would actually rename the component field. Special-casing that might be awkward though.

supersaiyansubtlety:
~~based on #340~~ rebased

Internal proposers, used to make several fixes and improvements:
- Enums:
  - [x] prevent renaming of enum constants (but still propose their names)
- Record components; #314
  - ~~propose getter names matching the corresponding field~~ already implemented
  - [x] prevent directly naming definite getters
  - [x] allow indirectly renaming the field corresponding to the getter via the getter instead
  - [x] give the getter the same highlighting as the field

<details>

<summary>No synthetic field/local proposals for now</summary>

These require more context from vineflower to reliably match corresponding fields/locals.

- Anomymous class synthetic fields corresponding to outer method params; #252
  - [ ] propose field names matching the corresponding param
  - [ ] allow indirectly renaming params corresponding to fields via the fields
  - [ ] give the field the same highlighting as the param

</details>


supersaiyansubtlety:
This is ready

I removed the synthetic field/local stuff because it requires more info from vineflower.

Also note that for records, "give the getter the same highlighting as the field" is *theoretically* true, but since #340 makes getter indexing match decompilers and decompilers hide definite getters, you never actually see it in the gui.
You *would* see it when using the BYTECODE decompiler, except `remapToNameable` [here](<https://github.com/QuiltMC/enigma/pull/333/changes#diff-6787fd875d17cbe4c860740e6989396bc7ccb204b4f02b3653532ceeeefd8128R62>) is false for BYTECODE.

PR #334, opened 2025-11-09
update dependencies for 2.7 and move to jspecify annotations

eristhea:
dependency updates:
- shadow: `8.1.1` -> `9.2.2` (move to gradleup version)
- guava: `33.2.1` -> `33.5.0`
- gson: `2.10.1` -> `2.13.2`
- asm: `9.7.1` -> `9.9`
- tinylog: `2.6.2` -> `2.7.0` (omg)
- vineflower: `1.11.0` -> `1.11.2`
- junit: `5.10.3` -> `6.0.1`
- jimfs: `1.3.0` -> `1.3.1`
- gradle: `8.7` -> `9.2.0`

new dependencies:
- jspecify: `1.0.0`

previously, we depended on one of our libraries shipping a copy of the javax annotations. that was removed, so we've chosen to move to the new jspecify annotations instead of adding a manual dependency on javax.

this PR also reorders dependencies using intellij's automated dependency optimisation and resolves a few deprecations related to the gradle update.

thanks @supersaiyansubtlety for redoing this after i inexplicably broke everything

### DIFF
--- a/enigma/build.gradle
+++ b/enigma/build.gradle
@@ -185,9 +185,9 @@ final testPreHandlingPluginAgainstCurrentEnigma = tasks
 
 	classpath(cliShadowJar, pluginJarFile)
 
-	errorOutput new ByteArrayOutputStream()
+	errorOutput = new ByteArrayOutputStream()
 
-	ignoreExitValue true
+	ignoreExitValue = true
 
 	doFirst {
 		args 'print-stats', pluginJarFile.get().asFile, pluginJarFile.get().asFile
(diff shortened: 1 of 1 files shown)


enigma-swing/build.gradle
```diff
@@ -3,9 +3,7 @@ import com.formdev.flatlaf.FlatDarkLaf
 import com.formdev.flatlaf.FlatLightLaf
 import quilt.internal.task.PrintColorKeyGroupsMapCodeTask
 
-import javax.swing.LookAndFeel
-import javax.swing.UIManager
-
+import javax.swing.*
```
supersaiyansubtlety: I'd like to avoid wildcard imports

supersaiyansubtlety (approved):
looks great!

enigma-swing/build.gradle
ix0rai: good catch!

PR #338, opened 2025-11-16
Menu search

supersaiyansubtlety:
aka 'Fun with data structures'

Adds a "Search menus" menu under the "Help" menu.

Tasks
- [x] make all current menus searchable (except Help and contents)
- [x] implement search aliases (search "debug" to find the `Dev` menu)
- [x] make ~~all~~ (most ~~?~~) current menu items searchable
- [x] contains string search
- [x] display search alias next to result
- [x] make clicking on a search result effectively click the corresponding element
- [x] hot keys
- ~~make selecting a search result also select the corresponding element (if swing allows it)~~
swing does not allow it; instead, hold shift to select the searched item

<details>

<summary></summary>

Yes, the abstraction behind tries is a 'bit' overkill.
Yes, the generics in `CompositeStringMultiTrie`* got a little out of hand.
Yes, it was fun to write.

</details>


supersaiyansubtlety:
ready!

enigma-swing/src/main/java/org/quiltmc/enigma/gui/element/menu_bar/SearchMenusMenu.java
```diff
@@ -300,7 +300,7 @@ private void refreshPopup() {
 
 			// HACK: re-show if shrinking to move the popup back down in case it had to be shifted up to fit items
 			// re-showing can also result in dropped keystrokes; do so as infrequently as possible
-			// note: the initial showing from JMenu would cause an SOE if we also showed here for the initial showing
```
eristhea: this is actually the correct version, for acronyms the use of 'a' or 'an' is based on pronunciation, and SOE starts with `ess`

supersaiyansubtlety: ah thanks, good to know
wasn't aware there was a convention

enigma-swing/src/main/java/org/quiltmc/enigma/gui/element/menu_bar/SearchMenusMenu.java
```diff
@@ -0,0 +1,653 @@
...
+				this.addActionListener(e -> {
+					clearSelectionAndChoose(Result.this.searchable, MenuSelectionManager.defaultManager());
+				});
+
+				this.searchablePath = buildPathTo(this.getSearchable()).unwrapOrElse(err -> {
+					Logger.error(err);
+					return ImmutableList.of();
+				});
+
+				this.setToolTipText(this.searchablePath.isEmpty() ? "????" : this.searchablePath.stream()
```
eristhea, line 411: love this 😭

enigma-swing/src/main/java/org/quiltmc/enigma/gui/element/PlaceheldTextField.java
```diff
@@ -0,0 +1,198 @@
...
+import java.awt.Graphics;
+import java.awt.Insets;
+
+import static org.quiltmc.enigma.gui.util.GuiUtil.getCenteredFontBaseY;
+import static org.quiltmc.enigma.gui.util.GuiUtil.trySetRenderingHints;
+
+/**
+ * A text field that displays placeholder text when it's empty.
+ */
+public class PlaceheldTextField extends JTextField {
```
eristhea, line 20: this is a really useful component, excited to have it :)

enigma/src/main/java/org/quiltmc/enigma/util/StringLookup.java
```diff
@@ -0,0 +1,459 @@
...
+			StringMultiTrie<ResultWrapper<R>> resultsBySubstring
+	) {
+		this.substringDepth = substringDepth;
+		this.comparator = comparator;
+
+		this.resultsByPrefix = resultsByPrefix;
+		this.resultsBySubstring = resultsBySubstring;
+	}
+
+	// TODO implement lookUpIgnoreCase
```
eristhea, line 159: looks like lookup is already case insensitive, this can be removed

```suggestion
```

eristhea (approved):
pretty surface-level review as i only skimmed the data structure fun. i love this! so overengineered but just like the entry markers it's really really impressive how well this works. i especially like the hints in the search menu, really great way to teach the features

enigma/src/main/java/org/quiltmc/enigma/util/StringLookup.java
supersaiyansubtlety: Tries support case-insensitive queries, but `SearchMenusMenu`'s `StringLookup` fakes case-insensitive queries by storing everything in lowercase.
Currently you can't create a `StringLookup` that supports case-sensitive *and* case-insensitive queries.

PR #343, opened 2025-12-02
Redesign stats menu

### LINKED ISSUE #283: redesign mapping stats menu
this thing is simply awful.
Image

points that need to be addressed:
- it's completely incomprehensible how these little checkboxes next to the types factor in
  - do they mean the type is included in the overall percentage? are they used for the filter function?
  - instead of putting these in the main menu, there should probably be a sub-menu accessed via the 'generate diagram' button
- spacing is inconsistent -- look at where the slashes are placed in the different types
  - each line is set up in 4 right-aligned columns (number, /, number, percentage), which could be good but in practice looks weird
- no visualization of the data
- no indication of which settings are applied in the data being shown
- no delineation between the stats and the settings
- text is all the same font and size -- no data is highlighted compared to others

PiTheGuy:
Closes #283 

image


supersaiyansubtlety (changes requested):
The new dialog looks great, thanks for this!
It's much easier to get an idea of the state of things with the information presented visually.

Just a few nitpicks and some size/scale issues.

---

<details>

<summary>The settings button get misaligned if you resize the window</summary>

It should probably stay at the bottom, or disabling resizing is probably fine, too.

Image

</details>

---

<details>

<summary>Not everything scales</summary>

Scaling is missing for:
- the "Active Settings" label
- the overall percentage
- the corner radius of individual progress bars

| 50% | 100% | 200% |
| - | - | - |
| 50% scale dialog  | 100% scale dialog | 200% scale dialog |

</details>

---

supersaiyansubtlety (changes requested):
Found another issue with resizing (shrinking) the dialog:
enigma-343-small-dialog

Making the scroll pane fill `BOTH` and giving it non-zero y weight partially fixes it:
enigma-343-small-dialog-partial-fix

But it still has a vertical scroll bar despite having empty space below it; I think the overall stat bar circle is taking up space even though it's not being rendered.

supersaiyansubtlety (approved):

eristhea (approved):
lgtm! sorry for the long wait for the review

PR #351, opened 2025-12-14
Fix constructor inheritance

### LINKED ISSUE #349: Constructors resolve to super constructors
`IndexEntryResolver` resolves constructors to super-classes' constructors with matching params.

The issue probably originates in `InheritanceIndex` and is probably solvable with a simple `isConstructor()` check.

supersaiyansubtlety:
fixes #349

- adds a check to `IndexEntryResolver` to make constructors always resolve to themselves
- removes redundant `instanceof ClassEntry` checks that could never be `true` because `getClassChild` already excluded them


enigma/src/main/java/org/quiltmc/enigma/api/translation/mapping/IndexEntryResolver.java
```diff
@@ -73,8 +73,12 @@ public <E extends Entry<?>> Collection<E> resolveEntry(E entry, ResolutionStrate
 			}
 
 			// Don't search existing private and/or static entries up the hierarchy
-			// Fields and classes can't be redefined, don't search them up the hierarchy
-			if (access != null && (access.isPrivate() || access.isStatic() || entry instanceof FieldEntry || entry instanceof ClassEntry)) {
+			// Fields can't be redefined, don't search them up the hierarchy
+			// Constructors can't be redefined, don't search them up the hierarchy
+			if (
+					(access != null && (access.isPrivate() || access.isStatic() || entry instanceof FieldEntry))
```
supersaiyansubtlety: this only checks for `FieldEntry`s when `access` is non-null, as before
that doesn't seem correct, but changing it breaks tests

eristhea: cc @IotaBread 
reordering the check breaks shadow behaviour
i don't know exactly _when_ the access will be null -- if we can predict which fields will have their definitions indexed, we need to document our reliance on that behaviour, and if we can't, this needs to be changed

eristhea (approved):
besides the one quirk this looks great :) we can always look into that later

PR #352, opened 2025-12-17
Fix 348

### LINKED ISSUE #348: Naming a lib method param creates a lib mapping
ex: naming the param of an `Object::equals` implementation creates `java/lang/Object.mapping`

probably introduced in #306, specifically because it:
> removes `java/lang/Object`'s explicit exclusion from `InheritanceIndex`

supersaiyansubtlety:
Fixes #348

#306 made `EntryRemapper` use the combined `JarIndex` instead of the main `JarIndex`

That allowed its `MappingValidator` to detect conflicts with inherited lib methods, but resulted in its `EntryResolver` resolving roots in lib classes.

This makes `EntryRemapper` use the combined index for its `MappingValidator` and the main index for its `EntryResolver`.

This required breaking changes to `EntryRemapper`'s factory methods.


eristhea (approved):
looks great!

PR #353, opened 2025-12-17
Name via constructor

### LINKED ISSUE #350: Allow naming classes via their constructors
Currently `IdentifierPanel` doesn't allow renaming a class when the caret is in a constructor.

Allowing this should be possible using `EntryReference::getNameableEntry`.

supersaiyansubtlety:
- `IdentifierPanel`:
  -  checks the nameability of the cursor's reference instead of its reference's entry in `TableHelper`; fixes #350
  - removes redundant inner/outer class checks when getting `ClassEntry` names; `getName` already returns the simple same for inner classes and the full name for outer classes
- `EditorPopupMenu`:
  -  accounts for cursor state in `EditorPopupMenu::retranslateUi`
  - checks the nameability of the cursor's reference instead of its reference's entry when setting `toggleMappingItem`'s text; fixes the text always being "Mark as deobfuscated" for constructors; even when deobfuscated

### DIFF
--- a/enigma-swing/src/main/java/org/quiltmc/enigma/gui/Gui.java
+++ b/enigma-swing/src/main/java/org/quiltmc/enigma/gui/Gui.java
@@ -434,8 +434,8 @@ public class Gui {
 		editor.navigateToToken(tokens.get(0));
 	}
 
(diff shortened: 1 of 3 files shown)


eristhea (approved):
looks great :) i love little updates like this

PR #355, opened 2025-12-18
Inline syntaxpain

supersaiyansubtlety:
> [!CAUTION]
> Do not squash merge

closes https://github.com/QuiltMC/syntaxpain/issues/9

---

- inlines https://github.com/QuiltMC/syntaxpain with `includeBuild` in `syntaxpain/`; git history is maintained
- replaces the `org.xbib.gradle.plugin.jflex` plugin with a `jflex.anttask.JFlexTask` in build script; the plugin used deprecated gradle features, so the build workflow's `--warning-mode fail` would cause it to fail

---

once this is merged we should:
- transfer https://github.com/QuiltMC/enigma/issues/359
- archive https://github.com/QuiltMC/syntaxpain


eristhea (approved):
looks great! i skimmed through the syntaxpain code and. wow. why does every file have its own comment and code style

PR #357, opened 2025-12-25
Mostly fix `NetworkTest`

supersaiyansubtlety:
*Mostly* fixes `NetworkTest` flakiness.

Merry happy.

---

- waits for client connection in `testLogin` and `testTakenUsername` to avoid race conditions
- removes unnecessary `EnigmaClient::disconnect` calls after waiting on `DummyClientPacketHandler.disconnectFromServerLatch`
- kicks clients instead of directly disconnecting them when not waiting on the disconnection latch;
  fixes race condition where the `kick` calls of servers' client threads sent packets that were received by other tests' clients
- adds repetitions to most tests

---

> [!WARNING]
> `testTakenUsername` is still flaky on CI

I didn't add repetitions for `testTakenUsername` because it's still flaky on CI; I can't reproduce the failures locally.

I thought the issue might be that the build workflow runs on jdk 17+20 were interfering with each other, but it's still flaky when they're sequential (10,000 reps here):
- [sequential build](https://github.com/supersaiyansubtlety/enigma/blob/fix-NetworkTest-countdown-sequential-build/.github/workflows/build.yml#L11)
- [run failure](https://github.com/supersaiyansubtlety/enigma/actions/runs/20512753501/job/58935732156#step:4:225)

### DIFF
--- a/enigma-server/src/main/java/org/quiltmc/enigma/network/EnigmaServer.java
+++ b/enigma-server/src/main/java/org/quiltmc/enigma/network/EnigmaServer.java
@@ -87,7 +87,8 @@ public abstract class EnigmaServer {
 		thread.start();
 	}
 
-	private void acceptClient() throws IOException {
+	@VisibleForTesting
+	Socket acceptClient() throws IOException {
 		Socket client = this.socket.accept();
 		this.clients.add(client);
 		this.unapprovedClients.add(client);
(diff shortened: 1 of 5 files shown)


eristhea (approved):
looks great! excited to not be dealing with this on the ci literally every other second

PR #362, opened 2026-08-24
Repair inner class nesting when stripped by an obfuscator

RDIL:
There's a lot going on here, so I'm going to try and break it down as simply as I can:

- Enigma has two concepts of how nesting works which disagree with each other. (1) It grabbed it from the `$` in a class name, but (2) decompilers get it from the InnerClasses attribute.
- Some obfuscators (like the one used on Charles Proxy, not sure which it actually is) strip out this nesting attribute, which lead to phantom classes that Enigma could tell were nested, but the decompiler couldn't.
- This meant you couldn't edit them! (Or even see them in the GUI's source tree)

So, how have I gone about fixing this? Glad you asked!

Class trees, class search, `Gui.showReference`, `moveClassTree`, javadoc/stat invalidation (and many more) are now all aware of this and can handle appropriately. Most importantly, we stop silently dropping `$` classes.

A lovely new test has been added to prevent it from regressing (I hope)


enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/EnigmaTextTokenCollector.java
```diff
@@ -155,27 +172,9 @@ private void parseSource() {
 			this.addClassAndChildren(decl, pkgPrefix + decl.getNameAsString());
 		}
 
-		for (ClassEntry classEntry : this.classRanges.keySet()) {
-			String[] parts = classEntry.getContextualName().split("\\$");
```
RDIL, line 159: without this refactor, it tries to find an outer class with a named inner class that doesn't exist, and would throw.

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/EnigmaTextTokenCollector.java
```diff
@@ -425,7 +426,11 @@ private MethodEntry getSyntheticMethodEntry(SyntheticMethodSpan method) {
 	public void start(String content) {
 		this.content = content;
 		this.lineIndexer = new LineIndexer(content);
+		this.current = new ContentTokens();
+		this.currentIsKeyed = false;
 		this.classRanges.clear();
+		this.classDeclarations.clear();
+		this.classStack.clear();
```
RDIL, line 433: existing bug!: this wasn't cleared between runs, which could lead to a different file's tokens occasionally being leaked onto an unrelated file, in very particular circumstances

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/VineflowerSource.java
```diff
@@ -74,37 +78,49 @@ private void checkDecompiled() {
...
+		EnigmaResultSaver saver = new EnigmaResultSaver(this.index);
 		Map<String, Object> options = getOptions(new EnigmaJavadocProvider(this.remapper), this.settings);
 		IFernflowerLogger logger = new EnigmaFernflowerLogger();
 		BaseDecompiler decompiler = new BaseDecompiler(saver, options, logger);
 
-		AtomicReference<EnigmaTextTokenCollector> tokenCollector = new AtomicReference<>();
+		List<EnigmaTextTokenCollector> tokenCollectors = Collections.synchronizedList(new ArrayList<>());
 		TextTokenVisitor.addVisitor(next -> {
-			tokenCollector.set(new EnigmaTextTokenCollector(next));
-			return tokenCollector.get();
```
RDIL, line 85: this didn't work because `tokenCollector.get()` would only ever resolve the current file, even if we're asking for a different file!

enigma-swing/src/main/java/org/quiltmc/enigma/gui/node/ClassSelectorClassNode.java
```diff
@@ -122,7 +122,7 @@ public void done() {
 
 	@Override
 	public String toString() {
-		return this.deobfEntry.getSimpleName();
+		return this.deobfEntry.getContextualName();
```
supersaiyansubtlety, line 125: Is this change important?

enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/jar/InnerClassIndex.java
```diff
@@ -0,0 +1,61 @@
...
+
+	@Override
+	public void indexInnerClass(ClassDefEntry classEntry, @NonNull InnerClassData innerClassData) {
+		this.nestedInSource.add(new ClassEntry(innerClassData.name()));
+	}
+
+	/**
+	 * Returns whether {@code entry}'s source is written inside another class' source, i.e. whether the jar
+	 * declares it nested with an {@code InnerClasses} record. A class whose name looks nested but that has no
+	 * such record decompiles to a file of its own, and this returns {@code false} for it.
```
supersaiyansubtlety:
```suggestion
	 * declares it nested with an {@code InnerClasses} record. A class whose name looks nested but that has no
	 * such record yields {@code false}.
```
unless mentioning the separate file here is particularly important
removing that part makes it flow better imo

enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/jar/InnerClassIndex.java
```diff
@@ -0,0 +1,61 @@
...
+public class InnerClassIndex implements JarIndexer {
+	private final Set<ClassEntry> nestedInSource = new HashSet<>();
+
+	@Override
+	public void indexInnerClass(ClassDefEntry classEntry, @NonNull InnerClassData innerClassData) {
+		this.nestedInSource.add(new ClassEntry(innerClassData.name()));
+	}
+
+	/**
+	 * Returns whether {@code entry}'s source is written inside another class' source, i.e. whether the jar
```
supersaiyansubtlety:
```suggestion
	 * Returns whether the passed {@code entry}'s source is written inside another class' source, i.e. whether the jar
```
nitpick for consistent javadoc style

enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/jar/InnerClassIndex.java
```diff
@@ -0,0 +1,61 @@
...
+	 * such record decompiles to a file of its own, and this returns {@code false} for it.
+	 *
+	 * @param entry the class to check
+	 */
+	public boolean isNestedInSource(ClassEntry entry) {
+		return this.nestedInSource.contains(entry);
+	}
+
+	/**
+	 * Returns the class whose source contains {@code entry}'s, which is {@code entry} itself unless it is
```
supersaiyansubtlety:
```suggestion
	 * Returns the class whose source contains the passed {@code entry}'s, which is {@code entry} itself unless it is
```

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/EnigmaContextSource.java
```diff
@@ -40,14 +45,30 @@ private void collectClassNames() {
...
-			this.classNames.addAll(this.classProvider.getClasses(root).stream().filter(s -> s.contains("$")).toList());
+			this.classNames.addAll(this.classProvider.getClasses(root).stream()
+					.filter(s -> !s.equals(root) && this.isNestedInSource(s))
+					.toList());
+		}
+	}
+
+	private String sourceRoot(String className) {
+		String root = className;
+		while (root.lastIndexOf('$') > 0 && this.isNestedInSource(root)) {
```
supersaiyansubtlety: minor nitpick, but I'd prefer if the result of `lastIndexOf` were stored rather than calling the method twice

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/EnigmaTextTokenCollector.java
```diff
@@ -94,26 +102,35 @@ private Token getToken(TextRange range) {
...
 	private void addReference(Token token, Entry<?> entry, Entry<?> context) {
-		this.references.put(token, Pair.of(entry, context));
-		this.tokens.put(token, false);
+		this.current.references.put(token, Pair.of(entry, context));
+		this.current.tokens.put(token, false);
 	}
 
-	public void addTokensToIndex(SourceIndex index, UnaryOperator<Token> tokenProcessor) {
-		for (Token token : this.tokens.keySet()) {
+	public boolean hasTokensFor(String className) {
```
supersaiyansubtlety, line 114: can this be private?

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/EnigmaTextTokenCollector.java
```diff
@@ -441,6 +446,12 @@ public void visitClass(TextRange range, boolean declaration, String name) {
 		this.updateMethodStack(range);
 
 		if (declaration) {
+			if (!this.currentIsKeyed) {
```
supersaiyansubtlety, line 449: could `currentIsKeyed` be replaced with `tokensByClass.containsKey(name)`?

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/VineflowerSource.java
```diff
@@ -74,37 +78,49 @@ private void checkDecompiled() {
...
-		} else {
-			tokenCollector.get().addTokensToIndex(this.index, token -> token);
+		EnigmaTextTokenCollector tokenCollector = null;
+		synchronized (tokenCollectors) {
+			for (EnigmaTextTokenCollector collector : tokenCollectors) {
+				if (collector.hasTokensFor(this.className)) {
+					tokenCollector = collector;
+					break;
+				}
+			}
```
supersaiyansubtlety:
```suggestion
		EnigmaTextTokenCollector tokenCollector;
		synchronized (tokenCollectors) {
			tokenCollector = tokenCollectors.stream()
				.filter(collector -> collector.hasTokensFor(this.className))
				.findFirst()
				.orElse(null);
```
I like to avoid reassignment when possible, and I think this is a bit clearer.

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/VineflowerSource.java
```diff
@@ -74,37 +78,49 @@ private void checkDecompiled() {
...
 		}
-	}
 
-	private static void removePackageStatement(SourceIndex index, EnigmaTextTokenCollector tokenCollector) {
 		if (tokenCollector == null) {
-			throw new IllegalStateException("No token collector");
+			Logger.warn("No tokens were collected for {}", this.className);
```
supersaiyansubtlety, line 106: should this ever happen?

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/EnigmaContextSource.java
```diff
@@ -40,14 +45,30 @@ private void collectClassNames() {
...
 		if (!options.containsKey(IFernflowerPreferences.DECOMPILE_INNER)
 				|| "1".equals(options.get(IFernflowerPreferences.DECOMPILE_INNER))) {
-			this.classNames.addAll(this.classProvider.getClasses(root).stream().filter(s -> s.contains("$")).toList());
+			this.classNames.addAll(this.classProvider.getClasses(root).stream()
+					.filter(s -> !s.equals(root) && this.isNestedInSource(s))
+					.toList());
+		}
+	}
+
+	private String sourceRoot(String className) {
```
supersaiyansubtlety: name nitpick: `getSourceRoot`

enigma/src/test/java/org/quiltmc/enigma/TestFlatInnerClasses.java
```diff
@@ -0,0 +1,138 @@
...
+
+		try (JarOutputStream out = new JarOutputStream(Files.newOutputStream(jar))) {
+			write(out, OUTER, declareNesting);
+			write(out, INNER, declareNesting);
+		}
+
+		return jar;
+	}
+
+	private static void write(JarOutputStream out, ClassEntry entry, boolean declareNesting) throws IOException {
```
supersaiyansubtlety: test classes go in `enigma/src/test/java/org/quiltmc/enigma/input/`

you can control whether `InnerClasses` is retained using (or not) `-keepattributes` in a custom proguard config file in `enigma/src/test/resources/proguard-$name-test.conf`

(see `enigma/build.gradle`'s `registerTestJarTasks`)

supersaiyansubtlety (changes requested):
Overall looks great, thank you!

I made a few nitpicks, and:
- test classes are handled by gradle and get obfuscated by proguard, details my comment on new test
- this makes a breaking change (`VineflowerSource` constructor), so it should target develop/2.8

enigma-swing/src/main/java/org/quiltmc/enigma/gui/node/ClassSelectorClassNode.java
RDIL: yeah - `refreshClasses` now filters on `!isNestedInSource` instead of `!isInnerClass`, so flat inner classes appear as roots in the class tree.

before: `Inner`
after: `Outer$Inner`

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/EnigmaTextTokenCollector.java
RDIL: no - `VineflowerSource.checkDecompiled` calls it to pick the right collector out of the list. It could be package-private, but `addTokensToIndex` right below it is public for the same reason. i can drop to package-private if you'd like but if so I'd recommend doing it on both at once

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/EnigmaTextTokenCollector.java
RDIL: they aren't equivalent. the flag means "first declaration in this file", vs containsKey means "this name hasn't been seen in any file yet", so it would also key the nested declarations inside a file. `Foo`'s file would register under both `Foo` and `Foo$Bar`. then `hasTokensFor("Foo$Bar")` succeeds for a file that isn't `Foo$Bar`'s

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/EnigmaTextTokenCollector.java
supersaiyansubtlety: it's fine as-is, I think I just derped when I searched for usages

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/VineflowerSource.java
RDIL: from what I can tell:

- gui: no
- api: yes, if I pass in a nested class's name

this was a hard fail before this changeset, but I think having a file be viewable but unindexed is a better failure mode than the editor not showing anything. defer to you on the final decision

enigma/src/test/java/org/quiltmc/enigma/TestFlatInnerClasses.java
RDIL: applied in 4e79ec588dbdc614c5c98c974a463796ef30642c

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/EnigmaContextSource.java
RDIL: applied in 4e79ec588dbdc614c5c98c974a463796ef30642c

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/VineflowerSource.java
RDIL: applied in 4e79ec588dbdc614c5c98c974a463796ef30642c

enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/jar/InnerClassIndex.java
RDIL: applied in 4e79ec588dbdc614c5c98c974a463796ef30642c

enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/jar/InnerClassIndex.java
RDIL: applied in 4e79ec588dbdc614c5c98c974a463796ef30642c

enigma/src/main/java/org/quiltmc/enigma/api/analysis/index/jar/InnerClassIndex.java
RDIL: applied in 4e79ec588dbdc614c5c98c974a463796ef30642c

enigma/src/main/java/org/quiltmc/enigma/impl/source/vineflower/VineflowerSource.java
supersaiyansubtlety: I see, it's more likely than I initially thought. I think the warning is good.

supersaiyansubtlety (approved):
Awesome!

OroArmor (approved):

